PR #49, opened 2014-11-14
Ported Braille Fun activity 

24ark:
I was not able to find a suitable image for tux plane and the image behind BrailleChar item.


bdoin:
This activity has been integrated in the devel branch (on our KDE git). I did not many code changes. I removed the falling letter as it was a duplicate of the text on the plane. Maybe many little graphical improvements.

PR #51, opened 2015-01-02
Louis braille activity

24ark:
I have finished porting Louis-Braille Activity. I have also finished working on braille-fun activity. I have pushed these changes to different branches.


bdoin:
Hi, I tested your work and it's pretty well done. I will integrate your work asap.

PR #53, opened 2015-01-06
Louis braille activity

24ark:


bdoin:
Merged in the devel branch on KDE's git

PR #54, opened 2015-01-26
Added computer activity

aarsee:


bdoin:
Before being merged this activity requires a lot of work. Need to be made more generic and provide a way to connect items and see an animation or a feedback about the connection.

bdoin:
Thanks for your work on that. It is still not evolved enough to be merged. Feel free to continue refining the activity until it become useable with children.

PR #58, opened 2015-02-12
clickanddraw: Workaround for unclickable points

avtobiff:
It is not enough to just change z index for points in order to make them
clickable. The problem is that points later in the click index grab the
click instead of the current point to click.

As a workaround the pointIndexToClick comparison with (current?) index
is reversed, to show where the point is actually clickable.


bdoin:
Thanks for your patch, this is merged in KDE's git master branch

PR #60, opened 2015-02-12
Gletters update data and fix sublevel count

avtobiff:


src/activities/gletters/gletters.js
```diff
@@ -99,7 +99,7 @@ function initLevel() {
         // If "sublevels" length is not set in wordlist, use the words length
         maxSubLevel = level.words.length
     }
-    items.score.numberOfSubLevels = maxSubLevel;
+    items.score.numberOfSubLevels = maxSubLevel + 1;
```
bdoin: I don't agree here. It currently works fine, for 5 data we go from 1/1 to 5/5

avtobiff: On Tue, Feb 17, 2015 at 10:34 PM, Bruno Coudoin
<notifications@github.com> wrote:
>
> In src/activities/gletters/gletters.js:
>
> > @@ -99,7 +99,7 @@ function initLevel() {
> >          // If "sublevels" length is not set in wordlist, use the words length
> >          maxSubLevel = level.words.length
> >      }
> > -    items.score.numberOfSubLevels = maxSubLevel;
> > +    items.score.numberOfSubLevels = maxSubLevel + 1;
>
> I don't agree here. It currently works fine, for 5 data we go from 1/1 to 5/5

I found it confusing to still have data left to type when it showed
5/5, I assumed
I had typed everything when reaching 5/5 but there was one character left.

I always viewed the sublevel as a score count, which created the confusion.
Starting the count on zero (0) would better match this.

src/activities/gletters/resource/default-en.json
```diff
@@ -7,6 +7,16 @@
             "e",
             "i",
             "o",
+            "u",
```
bdoin: I understand what you want to do and it makes sense. But I don't like to to it by extendind the dataset this way. We have a property "sublevels" that can be used in the level. I tested it and it does not do the trick to enlarge the dataset up to this number. If you are interested, I propose you to change the code to allow this.

src/activities/gletters/gletters.js
bdoin: I agree with you, I am also confused by this but I have to switch my brain from software mode to user mode. A user expect the first level to be 1, not 0. And 5/5 means you at at the last level but not completed it.

bdoin:
Won't merge.

PR #61, opened 2015-02-15
adding intro_gravity

siddhism:
Complete port of intro_gravity. changed the reading behaviour of instructions with a button as suggested.


src/activities/intro_gravity/Intro_gravity.qml
```diff
@@ -0,0 +1,205 @@
...
+*   This program is distributed in the hope that it will be useful,
+*   but WITHOUT ANY WARRANTY; without even the implied warranty of
+*   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+*   GNU General Public License for more details.
+*
+*   You should have received a copy of the GNU General Public License
+*   along with this program; if not, see <http://www.gnu.org/licenses/>.
+*/
+import QtQuick 2.2
+import QtQuick.Controls 1.3
```
bdoin: Use 'import QtQuick.Controls 1.2' to accept older Qt 5.3 which we use on Android

bdoin:
When running it I see this log:
SliderStyle.qml:205:13: QML Loader: Binding loop detected for property "y"

When started in window mode, I don't see the 'let's' go button. You must take care your activity runs fine when you resize your window small or large. It is ok if you reset the level when the window is resized.
When I loose and I go to level 2 I am still in crash mode.
At level 2, probably others there is too many asteroids and they are too slow
The image background.svgz must be fully svg, not a png in a svg or we loose the advantage of the svg.

siddhism:
1. qt control 1.2 :done.
2. I couldn't figure how to correct it. there is a known bug.
https://bugreports.qt.io/browse/QTBUG-31441

bdoin:
Le 19/02/2015 05:27, siddhesh suthar a écrit :
> 1. qt control 1.2 :done.
> 2. I couldn't figure how to correct it. there is a known bug.
> https://bugreports.qt.io/browse/QTBUG-31441
>
The second issue here is on 'y' while the bug is on 'value', not sure 
this is the same issue as the bug you found.

siddhism:
3. Let's go button is now visible. fixed.
4. Loose and go to level 2 crashing problem fixed. everything works fine
now after resizing. fixed arrow's scale size according to background.width
5. background.png converted to .svg by using inkspace. fixed.

siddhism:
yes sir i need help with that issue.binding error are caused due to cyclic
dependencies but i don't see any now. i am pushing changes. everything else
seems to be fixed.

hkaelber:
Strange one, this loop! Seems to go away if you set the dimensions of your sliders, e.g.

width: 10
height: planetLeft.height

The Slider somehow seems to mess up its dimensions internally otherwise.

siddhism:
Increased speed of asteroids. can i destroy the asteroids as soon as they
cross the shuttle.

siddhism:
Great. got rid of the loop . thanks a lot.

PR #62, opened 2015-02-19
adding intro_gravity

siddhism:
1. removed binding loop for 'y'. 
2. let's go button visible on screen.
3. loose and go to level 2 crash problem fixed.
4. background image is now svg.
5. falling speed improved at different levels.


bdoin:
- converting an png to svg create extra useles layers and complexity. Please create one yourself, put a black background and add some random white points. Show us your artistic skills ;)
- Once lost there is no restart button
- The Asteroids should move by themselves with QML Animations intead of moving them with a timer.
- There is way too many asteroids to pass through.
- You should handle the resize case, catch the width or height change and reset the level.

siddhism:
I pushed a commit and covered the following.
1. created an svg using inkscape. somewhat artistic :)
2. created restart button. also clicking it on level on will directly start
without the instructions.Instruction are only showed when we start the game
first time.
3. Asteroid move done by fall duration.
4. only 5 asteroids per level implemented.

PR #63, opened 2015-02-20
adding intro_gravity

siddhism:
reset level on screen size change.created skip instruction and reload button definations different for first level and other levels.


bdoin:
You could place the help on the screen top, this solve the issue of smaller screens.
Instead of let's go, it should be next and let's go on the last help panel.

siddhism:
Help is placed at the top now. also the button's text it changed to next
and lets go for last help panel.
also added space shuttle for tux to join and win.

PR #64, opened 2015-02-22
Port of intro_gravity complete.

siddhism:
completed the port of intro_gravity. added shuttle. Instruction moved to top. Let's go button's name changed to next.


bdoin:
Merged in the devel branch of our KDE git.

PR #66, opened 2015-03-07
lang activity

siddhism:
basic structure of lang activity without the mini games


bdoin:
You need to make a clean PR with only your activity in it.

PR #67, opened 2015-03-07
Excercise

kj2015:


bdoin:
Thanks for your exercice. We can see you can manage to develop on GCompris.

PR #70, opened 2015-03-09
Tic Tac Toe activity

pulkitgenius:
Hello sir, I have made the changes you suggested. Overlapping problem has been solved and user interface is improved. Please give your reviews :)


bdoin:
Congratulation for your improvements.

I see some little details you should change:
- The level should start at one
- The level should not be present in 2 players mode
- In 2 player mode you should alternate the first player
- The 'play first' button is confusing, I would suggest to simply alternate the first player and remove the button. 

Next PR it is ready to be integrated ;)

PR #75, opened 2015-03-22
Water Cycle initial

sagaragarwal94:
I have only checked on full screen.Please review and for resolution independence,I am working on it.


bdoin:
Need more work to make it screen size independant.
Use Qml States to manage the differents states.

sagaragarwal94:
Thanks,Sir.I am working on the screen size independency.

PR #85, opened 2015-08-08
Hangman activity completed 

clock21am:
masking done 
properly rebased 
activity completed


clock21am:
hi made those changes which johnny and you had pointed please take a look

PR #87, opened 2015-08-30
Bug: userEntry max length is 3. 

rdejoux:
It is a problem in level 7 (with 500 in the center of the target) because the total can be bigger than 999. I change length test with 4.

Tiny changset, but my son found the problem this afternoon.

Very nice job with this new GCompris. Best regards.


bdoin:
Hi, thanks a lot for your patch. I just pushed a better one in master where I make the entry area being one more length than the result.

PR #90, opened 2015-10-17
Demo Activity: Shape Recognizing

mudithkr:
name of demo game is mudit_test


petitlapin:
Hi,
thanks for your work. A few (minor) remarks:
* no need of "!!" at the end of each text for level 2.
* If possible, the icon should be a svg.
* It misses a copyright/readme file to know the images licenses (where did you get them)
* You don't need the checkError.svg file.
* The "simple" shape images can be merged with the ones use in sudoku (I would think than we should set them in the shape game, and use them in sudoku).
* The size of the images still have an issue: if you use the keyboard arrows to navigate from top line to bottom, all the shapes shift vertically

JohnnyJ

mudithkr:
@petitlapin
regarding 5th point - should I use sudoku images?

petitlapin:
Hi,

for the images size, the issue is in FindIt.qml, the size of the items inside container depends on the number of elements on it which is weird (for example, if you put only 3 elements they are too big).
You can look at it if you want but it is not mandatory.

JohnnyJ

bdoin:
Thanks for your exercise, stay motivated and focused to bring your work to the quality level expected by GCompris.

PR #91, opened 2015-10-18
Exersise

ayushagrawal288:
Demo Activity for GCompris 

Kindly Review


petitlapin:
Hi,
as said on irc:
 "heigth" should be "height" in the data files.
The js file is not necessary.
The toopltip texts should be in qsTr("") for translation.
I'm not sure the intro message is necessary, but it should be generalized to all babymatch activities if it is.
I don't think it's necessary as it should not be hard to guess what to do (or clicking on the help button where the same info is).

However, it's a good use of shapegame activities :).

JohnnyJ

bdoin:
Thanks for your exercise, stay motivated and focused to bring your work to the quality level expected by GCompris.

PR #92, opened 2015-10-18
to merge Hangman

clock21am:


bdoin:
When I loose I have the bad bonus and the word is displayed but there is not normal way to go to the next word. Clicking on a char append it and go to the next word but this is not the best way.

When I click on the config button nothing happens and GCompris becomes locked. Can't do anything more.

clock21am:
about the bad bonus and the word display after the is for the kid to know what is the correct answer for this question 
and about the GCompris lock it is not locking in my system now yes but it did lock in my system for the frist time i build after rebasing 
should i not display the word after wrong trials ?

bdoin:
Yes, you should display the word but you should provide perhaps an OK button to move to the next word.

clock21am:
okay button is done and config freezing is also done :)

PR #95, opened 2015-10-21
 louis Braille video link modified

Nessa594:
earlier link was not working


bdoin:
Thanks, this is merged on master

PR #97, opened 2015-10-24
Fixes the `sed` compatibility issues with OSX.

yask123:


bdoin:
Thanks, this is merged

PR #100, opened 2015-11-04
Magic hat : Amélioration sur le comportement, ajout du bouton valider

Thibrom:


bdoin:
Thanks a lot, you patch is merged in master.

PR #101, opened 2015-11-26
Stefan.toncu, beginner test app 1.0

StefanToncu:
I followed these steps: [An exercise for new contributors]  (http://gcompris.net/wiki/An_exercise_for_new_contributors) and i created a new app called "Stefan.toncu" in which i added a simple button. Then i added an event: when the mouse hovers over the button, the button's color changes to red and when the mouse leaves the button's area, it changes back to its original color, white. When the button is clicked, it changes the text from "button label" to "Clicked!!"


bdoin:
Thanks for your exercise, this is a very good start. Now it's time to go further and create a real activity ;)

PR #104, opened 2015-12-15
Fix Animation of image transition

pulkitgenius:
This will fix the animation issue.

### DIFF
--- a/src/activities/babymatch/DragListItem.qml
+++ b/src/activities/babymatch/DragListItem.qml
@@ -62,7 +62,7 @@ Item {
         }
         onStopped: {
             tileImage.parent = tileImage.tileImageParent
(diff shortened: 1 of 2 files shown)


bdoin:
It is getting close but with large pieces they are drop on a near by spot instead of the one you are.

src/activities/babymatch/DragListItem.qml
```diff
@@ -266,9 +271,7 @@ Item {
     function updateOkButton() {
         if(view.areAllPlaced()) {
             showOk.start()
-        } else if(view.okShowed) {
```
pulkitgenius: Not needed as onPressed handle this automatically.

bdoin:
I think this one is not more relevant, many work was done to improve positioning lately.

PR #105, opened 2015-12-20
Portrait view for Chess

StefanToncu:
I added the Portrait View mode for Chess Activity. On portrait (mobile version), in French, the "swap" button ("Inverser le tableau") was covered by the chessboard, so i moved the buttons at the top of the board.
Here are some pics with before and after.

before

after


bdoin:
your patch is interesting and I will integrate it. Some remark:
- Change you commit name as 'Stefan Toncu' instead of StefanToncu because KDE requires it.
- When you make a Pull Request make a clean branch, in this one there is an activity trial and this is confusing.
- Formatting rules, this '(items.isPortrait==true)?1:3' should be written: 'items.isPortrait ? 1 : 3'

PR #106, opened 2016-01-02
fixed Colorize bug in Click_on_letter

StefanToncu:
In Click_on_letter, when pressing a wrong carriage, nothing happens, because the Colorize effect is not added. The effect was removed in this commit: [hexagon, fix colorize effect] (https://github.com/bdoin/GCompris-qt/commit/006d11ed9bb50590d9e373a9477f6d3d80d421c7) and i added it back so it works on Android as well as other distributions.
Now, when clicking on a bad carriage, it is colored red, as before the Colorize effect removal.
// in "ClickOnLetter.qml", there are some spaces introduced by my computer (i don't know how) and i could not remove them; i hope its ok still


bdoin:
Thanks, merged in master.

PR #107, opened 2016-01-05
Use imgWidth and imgHeight of images when defined in the dataset, and remove some unused variables.

pulkitgenius:
This will correct the image size problem after it is dropped (As in the case of Chronos level 4), and some unused variables are removed.


bdoin:
Merged, in master. Thanks.

PR #108, opened 2016-01-10
fixed "Bad" Voice ("Check Answer") after passing a sublevel in Algorithm

StefanToncu:
In Algorithm, after passing sublevels 1 and 2, "items.bonus.good("flower")" is showing the flower, because the sublevel was passed, but to prevent it from going to the next level, was added "items.bonus.isWin = false", which plays a "bad sound".
I changed the Bonus in Algorithm.qml to go to the next SubLevel, instead the next Level; passing to the next level is now handeled in the "nextSubLevel" function.
At the end of a sublevel is still displayed a flower and at the end of the levels is shown Tux, as before, but now, it plays good sounds at the passing of each sublevel.

### DIFF
--- a/src/activities/algorithm/Algorithm.qml
+++ b/src/activities/algorithm/Algorithm.qml
@@ -228,7 +228,7 @@ ActivityBase {
 
         Bonus {
             id: bonus
-            Component.onCompleted: win.connect(Activity.nextLevel)
+            Component.onCompleted: win.connect(Activity.nextSubLevel)
         }
 
         Score {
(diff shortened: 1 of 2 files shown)


bdoin:
I don't understand what this patch is about. The commit log talks about voice but this activity don't have any specific.

petitlapin:
Hi, if you play with sound, you'll notice that a "bad" (try again) voice is given when finishing a sublevel due to the Bonus (items.bonus.good("flower"); items.bonus.isWin = false).
The patch is to have a good bonus voice when finishing a sublevel.
.

StefanToncu:
I think i wasn't specific enough so i decided to record my screen, to show the differences between before and after my changes: https://youtu.be/4bwJE1s6zjg
As petitlapin commented, my patch is about fixing the bonus sound  "Check Answer" (.isWin = false), that should not have been played after passing the sublevels.

bdoin:
Thanks for your excellent explanation. I integrated your work in master.

PR #109, opened 2016-01-11
Correct babymatch activities when GCompris window is resized

pulkitgenius:
This includes 
1) Missing of navigation buttons when number of items in a screen (nbItemsByGroup) changes due to resizing of the GCompris window.
2) Resizing of images when the GCompris window is resized.

### DIFF
--- a/src/activities/babymatch/DragListItem.qml
+++ b/src/activities/babymatch/DragListItem.qml
@@ -103,6 +103,16 @@ Item {
             property Item currentTargetSpot
             property bool pressedOnce
             property bool parentIsTile : parent == tile ? true : false
+            
+            onFullWidthChanged: correctDroppedImageSize()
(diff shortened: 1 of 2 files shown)


src/activities/babymatch/DragListItem.qml
```diff
@@ -103,6 +103,16 @@ Item {
             property Item currentTargetSpot
             property bool pressedOnce
             property bool parentIsTile : parent == tile ? true : false
+
```
bdoin: Here the line is empty but contains spaces. It is a good practice to not put space before an end of line.

bdoin:
In your commit log please prefix them by the module it refers like 'babymatch, commit description'. This makes it easier to read the log.

Your patch is in master now. Thanks.

PR #110, opened 2016-01-22
Searchfeature

rish9511:


src/activities/menu/Menu.qml
```diff
@@ -45,7 +47,8 @@ ActivityBase {
     id: menuActivity
     focus: true
     activityInfo: ActivityInfoTree.rootMenu
-
+    objectName: "menu"
+    signal textchanged(string message)
```
petitlapin: Use lower camel case for signals/methods/attributes (textChanged in this case)

src/activities/menu/Menu.qml
```diff
@@ -65,6 +68,8 @@ ActivityBase {
         }
     }
 
+
```
petitlapin: no need of these spaces

src/activities/menu/Menu.qml
```diff
@@ -114,6 +119,10 @@ ActivityBase {
             icon: menuActivity.url + "strategy.svg",
             tag: "strategy"
         },
+        {
+            icon: menuActivity.url + "search-icon.png",
```
petitlapin: use svg if possible (there are some in https://openclipart.org/)

src/activities/menu/Menu.qml
```diff
@@ -218,12 +235,31 @@ ActivityBase {
...
-                        particles.burst(10)
-                        ActivityInfoTree.filterByTag(modelData.tag)
-                        ActivityInfoTree.filterLockedActivities()
-                        ActivityInfoTree.filterEnabledActivities()
-                        menuActivity.currentTag = modelData.tag
-                        section.currentIndex = index
+
+
+                               if(modelData.tag === "search")
+                               {
```
petitlapin: the { should be on the same line as the if()

src/activities/menu/Menu.qml
```diff
@@ -450,6 +498,106 @@ ActivityBase {
...
+
+            visible:  false
+            anchors
+            {
+
+               bottom :
+               {
+                   if(horizontal)
+                       return activitiesGrid.top
+
```
petitlapin: clean empty lines, no need to have this much

src/activities/menu/Menu.qml
```diff
@@ -450,6 +498,106 @@ ActivityBase {
...
+                font.pixelSize: 28
+                font.bold: true
+                font.family: gctext.fontfamily
+                opacity: 0.5
+                onTextChanged:
+                {
+                    menuActivity.textchanged(input.text)
+                }
+
+                GCText
```
petitlapin: Is it used?

src/activities/menu/Menu.qml
```diff
@@ -450,6 +498,106 @@ ActivityBase {
...
+
+            TextField
+            {
+                id : input
+                anchors.fill: parent
+                text: ""
+                textColor: "black"
+                font.pixelSize: 28
+                font.bold: true
+                font.family: gctext.fontfamily
```
petitlapin: We should have a better way to access the font without using a unused text field. Maybe doing a GCTextField?

src/core/ActivityBase.qml
```diff
@@ -223,4 +223,9 @@ Item {
         anchors.fill: parent
         active: !activityInfo.demo && ApplicationSettings.isDemoMode
     }
+
```
petitlapin: no need of these spaces, this file shouldn't be updated

src/core/ActivityInfoTree.cpp
```diff
@@ -272,4 +275,73 @@ void ActivityInfoTree::init()
     qmlRegisterSingletonType<QObject>("GCompris", 1, 0, "ActivityInfoTree", menuTreeProvider);
     qmlRegisterType<ActivityInfo>("GCompris", 1, 0, "ActivityInfo");
 }
+void ActivityInfoTree::beginsearch()
```
petitlapin: camel case for methods

src/core/ActivityInfoTree.cpp
```diff
@@ -272,4 +275,73 @@ void ActivityInfoTree::init()
...
+{
+
+     m_menuTree = m_searched;
+     emit menuTreeChanged();
+     m_mainWindow = ApplicationInfo::getWindow()->findChild<QObject*>("menu");
```
petitlapin: You should do like for the filterByTag, call the method display(QString) in qml.

src/core/ActivityInfoTree.cpp
```diff
@@ -272,4 +275,73 @@ void ActivityInfoTree::init()
...
+     QObject::connect(m_mainWindow,SIGNAL(textchanged(QString)),this,SLOT(display(QString)));
 
+}
+void ActivityInfoTree::display(QString text)
```
petitlapin: use a const reference if possible

src/core/ActivityInfoTree.cpp
```diff
@@ -272,4 +275,73 @@ void ActivityInfoTree::init()
...
+       // templength = length of string that was there in the textfield before the present string
+       // if user erases a character from a string ,the search needs to be performed on all the activities
+       if(length < templength)
+       {
+
+           m_menuTreeFulltemp = m_menuTreeFull;
+           end = m_menuTreeFulltemp.length();
+       }
+
+       if(text.trimmed() == "")
```
petitlapin: you can use empty() method

src/core/ActivityInfoTree.cpp
```diff
@@ -272,4 +275,73 @@ void ActivityInfoTree::init()
...
+       {
+           m_menuTree.clear();
+           m_menuTreeFulltemp = m_menuTreeFull;
+           templength = 0;
+           emit menuTreeChanged();
+       }
+       else
+       {
+
+           while(start != end )
```
petitlapin: you can probably use a method in <algorithm> (remove_if for example) with a lamdba.

src/core/ActivityInfoTree.cpp
```diff
@@ -272,4 +275,73 @@ void ActivityInfoTree::init()
...
+           {
+               // if there is any activity that matches the description push it into the m_menutree
+               if(m_menuTreeFulltemp.at(start)->title().contains(text.trimmed(),Qt::CaseInsensitive))
```
petitlapin: We should filter on title, but also on description and maybe more

src/core/ActivityInfoTree.h
```diff
@@ -59,12 +61,17 @@ protected slots:
 private:
     // this is the full activity list, it never changes
     QList<ActivityInfo *> m_menuTreeFull;
+    QList<ActivityInfo *> m_menuTreeFulltemp;
```
petitlapin: do you really need 2 new attributes to store the activity lists?

src/core/ActivityInfoTree.h
```diff
@@ -59,12 +61,17 @@ protected slots:
...
     // represents the Menu view and can be filtered
     QList<ActivityInfo *> m_menuTree;
+    QList<ActivityInfo *> m_searched;
     ActivityInfo *m_rootMenu;
     ActivityInfo *m_currentActivity;
     static int menuTreeCount(QQmlListProperty<ActivityInfo> *property);
     static ActivityInfo *menuTreeAt(QQmlListProperty<ActivityInfo> *property, int index);
+    QObject *m_mainWindow;
+    int templength = 0;
+    QString searchedtext;
```
petitlapin: Same here, I'm not sure these 3 variables are useful

src/core/ApplicationInfo.cpp
```diff
@@ -197,6 +197,10 @@ void ApplicationInfo::setWindow(QQuickWindow *window)
 {
     m_window = window;
 }
+QQuickWindow* ApplicationInfo::getWindow()
```
petitlapin: not needed if you connect the signal/slot in qml

src/core/GCText.qml
```diff
@@ -124,6 +124,8 @@ Text {
      */
     readonly property int pointSize: font.pointSize
 
+    property string  fontfamily:GCSingletonFontLoader.fontLoader.name
```
petitlapin: Can't you use directly GCSingletonFontLoader.fontLoader.name for the TextField?

src/activities/menu/Menu.qml
```diff
@@ -153,7 +162,7 @@ ActivityBase {
         property int sectionCellHeight: sectionIconHeight * 1.1
 
         property var currentActiveGrid: activitiesGrid
-        property bool keyboardMode: false
+        property bool keyboardMode: false
```
petitlapin: no need of the space at the end

src/activities/menu/Menu.qml
```diff
@@ -169,7 +178,7 @@ ActivityBase {
         }
         Keys.onTabPressed: currentActiveGrid = ((currentActiveGrid == activitiesGrid) ?
                                                     section : activitiesGrid);
-        Keys.onEnterPressed: if(currentActiveGrid.currentItem) currentActiveGrid.currentItem.selectCurrentItem();
+        Keys.onEnterPressed:if(currentActiveGrid.currentItem) currentActiveGrid.currentItem.selectCurrentItem();
```
petitlapin: can you revert your change here (adding the space after ":")

src/activities/menu/Menu.qml
```diff
@@ -180,14 +189,21 @@ ActivityBase {
             id: section
             model: sections
             width: horizontal ? main.width : sectionCellWidth
-            height: horizontal ? sectionCellHeight : main.height - bar.height
+            height:
+            {
+               if(currentTag === "search")
```
petitlapin: the { should be on the same line as the if()

src/activities/menu/Menu.qml
```diff
@@ -218,12 +235,31 @@ ActivityBase {
...
-                        ActivityInfoTree.filterEnabledActivities()
-                        menuActivity.currentTag = modelData.tag
-                        section.currentIndex = index
+
+
+                               if(modelData.tag === "search")
+                               {
+                                   section.currentIndex = index
+                                   menuActivity.currentTag = modelData.tag
+                                   searchbar.visible = true;
```
petitlapin: if possible, set the variables directly in qml instead of using js. For example, in this case, in searchBar item you should have:
visible: modelData.tag === "search". This way, it will automatically update

src/activities/menu/Menu.qml
```diff
@@ -218,12 +235,31 @@ ActivityBase {
...
+                               else
+                               {
+                                   particles.burst(10)
+                                   ActivityInfoTree.filterByTag(modelData.tag)
+                                   ActivityInfoTree.filterLockedActivities()
+                                   ActivityInfoTree.filterEnabledActivities()
+                                   menuActivity.currentTag = modelData.tag
+                                   section.currentIndex = index
+                                   searchbar.visible = false
+                                   if(ActivityInfoTree.menuTree.length === 0 && modelData.tag === "favorite")
```
petitlapin: you should not have this if there

src/activities/menu/Menu.qml
```diff
@@ -218,12 +235,31 @@ ActivityBase {
...
-                        menuActivity.currentTag = modelData.tag
-                        section.currentIndex = index
+
+
+                               if(modelData.tag === "search")
+                               {
+                                   section.currentIndex = index
+                                   menuActivity.currentTag = modelData.tag
+                                   searchbar.visible = true;
+                                   warningOverlay.active = false;
```
petitlapin: you should not set it there

petitlapin:
Hi, I did a code review (for formatting I didn't write a note everywhere but you can check all the files you updated). I didn't test yet.

JohnnyJ

src/activities/menu/Menu.qml
rish9511: I  found this on clipart https://openclipart.org/detail/181256/search-icons  ,is this good?
and is there a problem with png format

src/core/ActivityInfoTree.cpp
rish9511: I think most people remember activities by their title name so it would be better if we just apply the title filter ,moreover applying other filters will the make feature slow .

src/core/ActivityInfoTree.h
rish9511: yes , m_menuTreeFulltemp is needed since we can't change m_menuTreeFull.

src/core/ActivityInfoTree.h
rish9511: m_mainWindow won't be needed if  I call the display function from menu.qml just like filterbytag and other functions but the other two are needed

src/core/GCText.qml
rish9511: yes forgot to make this change .will do this

src/activities/menu/Menu.qml
petitlapin: Good for me. We try to avoid png because of resizing (vector/bitmap).

rish9511:
Hi,
about the points you have mentioned above

1. Displaying all the activities by default - I am not sure about this one .I as a user would not want any activities to be displayed  by default although it might of  help for the younger ones .

2.I couldn't figure out how to get rid of this error ,will try again .

3. Agreed, could you please suggest some color combinations ?

4.I'll fix this 

5.Ya,will look nice 

6. It gets displayed automatically on my android device ,why do we need to do this explicitly ?

There's one thing I would like to want to add to the existing feature(just for android)
The activities should be displayed when the user presses the enter button(the tick button on virtual keyboard) ,at present if the user does that the first activity of all the searched activities starts .I think this is because activities grid has the focus while the user types .So at present user will have to press the back button of the virtual keyboard to see the activities matching the search term .Just try it on android and you will know and please provide feedback for that too ..

Thank you :)


Rishabh

src/core/ActivityInfoTree.h
petitlapin: But can't you use m_menuTree?

src/core/ActivityInfoTree.h
petitlapin: If I understand correctly the code, these 2 variables are used to "optimize" the search to only search in an already filtered tree when we add letters to the textField?
I'm not sure this optimization is needed, is it slow to always redo the computation?

petitlapin:
Hi,

for 1 let's wait other advices :)
for 2, I'll take a look later.
for 3, at least the text should be black and not grey. You can try with the gradient used in the GCDialog.qml.
for 6, it's not done in Desktop (I didn't look why). But if the user has explicitely asked in configuration to have a virtual keyboard to input text, we have to show him :).

I'll test later on android but as I understand the problem, there is not enough space to see the activities filtered because the virtual keyboard takes too much place and cover most of the activities. And we can't hide the virtual keyboard using Enter?
You can probably update Keys.onReturnPressed in Menu.qml to do what you want (unfocus the textField so it will hide the virtual keyboard if we are in the good tag?)?

Johnny

src/core/ActivityInfoTree.cpp
hkaelber: Tend to agree with rish9511: Title will probably be searched for most often. Not sure if also scanning description won't produce too many result. Might need to try it. Maybe configurable.

As a developper I'd really love to see the 'internal' name (activityInfo.name.split('/')[0]) searched because I best know this one, which is also not subject to translation. But that might not be a valid requirement :-) Or maybe also configurable?

hkaelber:
Hi,

some comments

1. Tend to agree with Johnny: If you don't type anything all activities should be presented because you have not yet defined any filtering constraint. Also it has the nice benefit that there would be again a screen with all activities visible (which was removed months ago, but useful e.g. if you want to scan the whole list rapidly for an icon you're searching for).

6. Let's distinguish the *system*-provided virtual keyboard (InputMethod on Android) from the virtual keyboard that we implemented ourselves in core/ (*gcompris*). Currently you're using the system's virtual keyboard, which is the only place in GCompris that it is in use. Not sure if we shouldn't use the *gcompris* virtual keyboard to always confront the children with the same tools.

Your last point (7.): Right, the space remaining in landscape mode is *really* limited, on my phone not even a single row of result is shown. Might be (little) more if using *gcompris'* virtual keyboard, also could try if it helps to make it semi-transparent.
Also, in my tests the search-results where not updated on every new letter but only on every completed word (after a blank).

Holger

src/core/ActivityInfoTree.cpp
bdoin: I agree with Holger. When doing support we need to be able to direct a
user to search the real name in english.

src/core/ActivityInfoTree.cpp
rish9511: empty and trimmed will behave same in this case

src/core/ActivityInfoTree.h
rish9511: these 2 variables are not used for optimizing .

rish9511:
@hkaelber 
On android(can't say about ios) as mentioned by JohnnyJ the virtual keyboard takes space and the  activities get displayed only if you press anything except the enter button .The incremental search keeps working in the background but displays the activities once you are done typing a word .

Sorry but I  could not understand your 2nd point .

src/core/ActivityInfoTree.cpp
petitlapin: I was saying empty() instead of the operator==, the trimmed will still be needed (text.trimmed().empty() instead of text.trimmed() == "").
It avoids a string creation, the comparison and is clearer to understand.

src/core/ActivityInfoTree.h
petitlapin: for me, the templength is used to store the length of the text already typed. And if we add a letter in the textfield, we don't do the search on all the activities but only the ones in the temp tree (which is the one filled by the previous search).
Also, if templength is the length of the previous text, it can be a local variable to the function (set before searchedtext = text;)
With your solution we have 4 trees (the full one, the one we display and two for optimizing the search time).
My question is: is it slow to always recompute the tree at each character input (basically using same code as filterByTag/filterLockedActivities), or do we need to optimize it? In both case, I'm not sure it's needed to add the 2 other trees as we could do the same with only m_menuTree (same code you did, we only store the current search text as attribute and if its length is bigger than the next search we recompute all else we can search inside the already filtered m_menuTree).

rish9511: Do we need to optimize it ? According to me yes.
I'll try to explain :
We have a total of more than 100 activities .The first time user types something in the text field the search is performed on those 100 or more activities ,now if we don't use optimization we keep performing the search on those 100 or more activities again and again  which  makes the feature a bit slow.I tested this using "time" library available in c++ to see how much time cpu takes for the computation ,with optimization it takes quite less time compared to a non-optimized code. 

According to what you suggested I have made some changes to the existing code .This is how it looks right now :
I am using m_menutree for filtering but we will still need one extra list(m_searched in this case).

Use  of "m_searched";
if the user changes the tab from search to some other tab without erasing the text  he searched then we need to display the activities   that were searched. Since the value of m_menutree gets changed on changing the tab  we need a temp list which will store the activities that were  searched 

void display(QString text)
{
       int length,start = 0 ;

    length = text.length();
    searchedtext = text;
    m_menuTree = m_searched;


    if(length < templength)
    {
        m_menuTree = m_menuTreeFull;

    }
    if(text.trimmed().isEmpty())
    {
        m_menuTree = m_menuTreeFull;
        templength = 0 ;

    }
    else
    {

        if(length != templength)
        {
            qDebug()<<endl<<"hey";
            while(start != m_menuTree.length())
            {
                if(!m_menuTree.at(start)->title().contains(text.trimmed(),Qt::CaseInsensitive))
                {
                    m_menuTree.removeAt(start);
                    start = start - 1;

                }
                start ++;

            }
        }
        templength = length;

    }
    m_searched = m_menuTree;
    filterEnabledActivities();
    filterLockedActivities();
    sortByDifficulty();
    emit menuTreeChanged();

}

bdoin: Can you give us the result of the time calculation so that we can take good decisions.

hkaelber:
Would be good to have the search-results updated onKeypress also on Android.

2. Virtual keyboard: In <gcompris>/src/core/VirtualKeyboard.qml we define a VirtualKeyboard element, that is used on mobile platforms in activities that need a virtual keyboard. Look at gletters or wordsgame e.g. That means we don't use the virtual keyboard provided by the mobile platform OS at all currently. Using it for search means that we add another input method that children need to cope with.

Using our/gcompris' own VirtualKeyboad for the search feature has pros (+) and cons (-):
+ Only one input method in all of gcompris
+ Can easily react onKeypress()
+ Could make it semi-transparent (useful?)

- We'd need a keyboard-layout for each supported language. This could be generated out of all available characters on the fly (as is done in various activities) or we'd need a standard layout (qwerty etc.) to be provided by the translators.
- The OS-provided virtual keyboard has layout and standard-features (like shift etc) out of the box.

Opinions? Bruno?

bdoin:
Sorry to be late to the game, I would like to propose a different approach.

My idea is to create a dictionary index upfront and use it for the search. For each word we keep a list of activities it appears in and in which field (the field is interesting to boost a match in a section or a title). When we build the index, we do the lowercase and we remove small words (may be a param for each language) but 1 and 2 letters are meaningless. We end up with a shorter index to search than doing full text each time. What we have to do is iterate the index (so O(n)) with a starts_with on each key. And we find all the relevant activity with a score easy to calculate. We then present the activity ordered by best score.

This index maybe computed either at startup and on language change or on first hit in the search depending on how long it takes.

src/core/ActivityInfoTree.h
rish9511: ex: while searching for tux 
time taken by cpu in micro-seconds
for:
t = 78
u = 82 
x= 20 
I don't have the code I was using before ,it used to give more convincing results(right now "u" is taking more time ,this shouldn't be happening but it's most probably because of time taken by other threads which are working simultaneously).

My  humble opinion is to use  optimization since its doesn't come with any bad consequence.I understand it's just 100 activities and even if we don't use optimization  time taken by cpu will be in micro-seconds which is not much  .

To me it's better to loop for less number of time on every iteration rather than looping for same number of time .

Correct me if am wrong anywhere :)

rish9511:
@bdoin 


Could you please give an example so that  I understand your approach better ?Sounds interesting though


From what  I have understood :
1.there will be no incremental search now
2. we store all the activities with a matching word in a particular index using the  map data structure ?

bdoin:
Let's say we score 10 for a title match and 20 for a section match.
Let's say we have:
Activity: 1, title: Tux reads, section: reading
Activity: 2, title: Tux counts, section: math
We create the index using a QMap (could also use a QMultiMap):
tux: {Activity: 1, score: 10}, {Activity: 2, score 10}
tux: {Activity: 1, score: 10}, {Activity: 2, score 10}
reads: {Activity: 1, score: 10}
counts: {Activity: 2, score: 10}

If we search 'tu' we can "grep" it in the index. Here we get tux and we display Activity 1 and 2 sorted by score. If we search 'tux read', we can sum the score of Activity 1 in index 'tux' and index 'read' so 40 and show activity 2 after it with a score of 20.

When we create the index we remove the single and two chars length words except for chinese type languages of course.

src/core/ActivityInfoTree.h
petitlapin: Premature optimization is the root of all evil :); If the issue is in micro seconds and not at least in milliseconds, I don't think it's useful.
There are other "optimizations" that can be done without removing code readability: using const QString & for the parameter, using std algorithms: erase, remove_if.
The loop is done for in the worst case 100 times (the number of activities) and only on user inputs. It's not a lot and probably not the worst part of time used. As you said, for the 'u', it took more time than for the 't'. So I agree with you that less loops are better in theory but in this case, I think it's not useful to lose in code readability to earn a few micro seconds that won't be visible by the user.

bdoin:
It is true that this need to work on mobile. I would stick to our own
keeboard populated dynamically. It will always be more children friendly
than a real mobile virtual keyboard. We also don't need capitals in this
search.

src/core/ActivityInfoTree.h
rish9511: @petitlapin 
I don't  understand why the cpu time is more for 'U' but yes I agree with you we are not gaining much even by optimization .  
Just want to know your opinion on :
1.What about the code I have given above ?
2 .Are we loosing code readability in the above code ?

Also ,just for  knowledge 
1. How using const optimizes the code?

I'll try and use erase and remove_if

petitlapin: for me we can still gain in readabily using standard algorithms.
If we have to keep the search text (I may be missing a case where we need it outside the method but for me it's not useful), we can probably keep the optimisation doing something like (if we don't need to store the search text, we can remove the if condition and always set m_Tree = m_TreeFull):
```c++
void display(const QString &newText) {
   if(newText.trimmed().length() < m_searchText.length()) m_Tree = m_TreeFull;
   if(!newText.trimmed().empty()) std::erase(std::remove_if( some code with m_Tree )); // I agree with the fact it's useless here to  compute for nothing ;)
   m_searchText = newText.trimmed();
   ... // the remaining code is good
}
```

-> this way, you don't manually iterate in the tree and don't have to handle the indices.

Also, for the m_search tree, I think we can recompute it when changing because we don't empty the text field when changing tab (from the search to another)?

The const probably won't make it faster but it helps for developers (we know that the string will not be updated in this method). Using a const reference will avoid to have to recreate a new QString

rish9511:
@bdoin

Sorry,I tried but failed to  understand your new approach . @petitlapin @hkaelber   would you guys help me out here ?

petitlapin:
As I understand, the approach bdoin proposes is to sort the activities depending on the words input in the search . So all the activities will always be displayed but in the order matching the best the search.
To take the example bdoin wrote, if we say every word in the title is worth 10 points and in the section 20 points, we have 2 activities:
Activity 1: title: Tux reads, section: reading
Activity 2: title: Tux counts, section: math
If the user inputs "read" then Activity 1 will have a score of 30 (10 because in the title + 20 in the section) and Activity 2 will have a score of 0 (because "read" in neither in title nor section) => Activity 1 will be shown before Activity 2.
If the user inputs "math", then Activity 1 will have a score of 0 (because counts is neither on title nor section) and Activity 2 will have a score of 20 (because math is in the section) => Activity 2 will be shown before Activity 1.
If the user inputs "Tux", then Activity 1 and 2 will have a score of 10 (because Tux is in both title) => same score, we can sort using the title for example.

Then, the thing with the map is how we could code it. If I understand, we split each activity titles and section, and for each word, we have a pair {activityName, score}. Then, to sort, we do a qSort() of the list with a sorting function computing each score for each activity.

If we do this approach we should also have a high score for the internal name of the activity (the one developers use).

I don't know which approach is better. The filtering method (what you did) is easier to code (already done :)). I think that if we don't erase after the remove_if, all the activities will still be displayed, it will be more binary: first the activities that match the search, then the others.
The approach bdoin suggests will add a granularity on the search results: some activities may match more or less the search so we will display them by order of matching.

I hope it's clearer.

bdoin:
My proposal came after it was mentioned performance issues. Based on your numbers if we talk microseconds we are just fine with your approach. You could even add a by field scoring with your approach if we want to go this way. So by now if performance is fine let's stick to your implementation. We may decide to improve the algo later on if needed.

PR #111, opened 2016-02-09
Searchfeature

rish9511:
I had to close the last PR since it had many conflicts .
I hope I have made all the changes that were discussed in the last PR and i know there are still some white space changes left but I tried to remove as many as i could:P


src/activities/menu/Menu.qml
```diff
@@ -45,7 +47,7 @@ ActivityBase {
     id: menuActivity
     focus: true
     activityInfo: ActivityInfoTree.rootMenu
-
+    signal textchanged(string message)
```
petitlapin: Is it used? If used, can you rename it to textChanged (else remove it :))

src/activities/menu/Menu.qml
```diff
@@ -122,6 +128,9 @@ ActivityBase {
         id: background
         source: menuActivity.url + "background.svg"
         sourceSize.width: parent.width
+        height: menuActivity.currentTag === "search" && ApplicationSettings.isVirtualKeyboard ?
+                    main.height - (((keyboard.keyHeight+keyboard.rowSpacing) * keyboard.rows)+keyboard.margin)
```
petitlapin: getting the keyboard.height is not the same? If not, there should be a property in it to get the value, it should not be computed outside

src/activities/menu/Menu.qml
```diff
@@ -218,12 +236,24 @@ ActivityBase {
...
-                        ActivityInfoTree.filterEnabledActivities()
-                        menuActivity.currentTag = modelData.tag
-                        section.currentIndex = index
+
+
+                               if(modelData.tag === "search"){
+
+                                   section.currentIndex = index
+                                   menuActivity.currentTag = modelData.tag
+                                   ActivityInfoTree.display(input.text);
```
petitlapin: I would rename display to filterBySearch or something like that. display is too generic

src/activities/menu/Menu.qml
```diff
@@ -459,6 +501,148 @@ ActivityBase {
...
+                id : input
+                anchors.fill: parent
+                text: ""
+                textColor: "black"
+                font.pixelSize: 28
+                font.bold: true
+                horizontalAlignment: TextInput.AlignHCenter
+                verticalAlignment: TextInput.AlignVCenter
+                font.family:GCSingletonFontLoader.fontLoader.name
+                activeFocusOnPress:  ApplicationSettings.isVirtualKeyboard ? false: true
```
petitlapin: can be simplified to !ApplicationSettings.isVirtualKeyboard

src/activities/menu/Menu.qml
```diff
@@ -459,6 +501,148 @@ ActivityBase {
...
+                onTextChanged:
+                {
+                    ActivityInfoTree.display(input.text)
+                }
+
+                GCText
```
petitlapin: unused

src/activities/menu/Menu.qml
```diff
@@ -459,6 +501,148 @@ ActivityBase {
...
+            anchors
+            {
+                top: background.bottom
+                left:  parent.left
+             }
+            onKeypress:
+            {
+                processKeyPress(text);
+
+                function processKeyPress(text_){
```
petitlapin: if it is the processKeyPress called above, no need of the function

src/core/ActivityInfoTree.cpp
```diff
@@ -283,5 +285,66 @@ void ActivityInfoTree::init()
...
+            m_searched = m_menuTreeFull;
+        }
+
+        m_menuTree = m_menuTreeFull;
+    }
+
+
+
+    if(!text.trimmed().isEmpty() and text.length() != m_searchedtext.length())
+        m_menuTree.erase(std::remove_if(m_menuTree.begin(),m_menuTree.end(),
```
petitlapin: as discussed in the other PR, it would be great to be able to also search on the internal name, can you add it?

src/activities/menu/Menu.qml
rish9511: Why?

src/activities/menu/Menu.qml
rish9511: Sorry but didn't understand your comment very well.

petitlapin: in your case, you compute the height of the virtual keyboard. It should be a property of the virtualKeyboard.
Instead of ((keyboard.keyHeight+keyboard.rowSpacing) * keyboard.rows)+keyboard.margin), you should have something like keyboard.height

src/activities/menu/Menu.qml
petitlapin: in this case, it's not called anywhere else so it's not useful

src/activities/menu/Menu.qml
```diff
@@ -459,6 +500,143 @@ ActivityBase {
...
+        }
+
+        VirtualKeyboard
+        {
+            id:keyboard
+            parent: background
+            property string search;
+            property var letter;
+            property int rows;
+            width: horizontal ? main.width :parent.width - (section.width +10)
```
rish9511: i have set the height here ,just before width 
height:((keyHeight + rowSpacing) * 3)+ margin

petitlapin:
A few global points:
* If I have the virtual keyboard displayed, I can't input letters with real keyboard.
* When I resize only horizontally, I have: qrc:/gcompris/src/activities/menu/Menu.qml:504:9: QML Rectangle: Possible anchor loop detected on vertical anchor.
* In vertical mode (sections are on left), the text is not displayed inside the search rectangle (more visible when the virtual keyboard is displayed).

More for discussion than for fix:
* in vertical mode, with keyboard, the section icons are too small because there is not a lot of room, how can we improve this? Increase the sectionBar width and do 2 rows of icons?
* with keyboard, in both horizontal and vertical mode, we don't see activities completely.

Johnny

src/activities/menu/Menu.qml
```diff
@@ -218,12 +235,24 @@ ActivityBase {
...
                     function selectCurrentItem() {
-                        particles.burst(10)
-                        ActivityInfoTree.filterByTag(modelData.tag)
-                        ActivityInfoTree.filterLockedActivities()
-                        ActivityInfoTree.filterEnabledActivities()
-                        menuActivity.currentTag = modelData.tag
-                        section.currentIndex = index
+
+
+                               if(modelData.tag === "search"){
```
petitlapin: When I change the difficulty, all the activities are displayed by default, that should not be the case, only the activities with the selected difficulties should be displayed.
The code to do this is currently in filterByTag. I would put the common code in a new method filterByDifficulty(int levelMin, int levelMax) and call it from both filterByTag() and filterBySearch().

src/activities/menu/Menu.qml
```diff
@@ -504,9 +675,11 @@ ActivityBase {
...
         }
         onClose: {
-            ActivityInfoTree.filterByTag(menuActivity.currentTag)
-            ActivityInfoTree.filterLockedActivities()
-            ActivityInfoTree.filterEnabledActivities()
+            if(menuActivity.currentTag != "search"){
+                ActivityInfoTree.filterByTag(menuActivity.currentTag)
+                ActivityInfoTree.filterLockedActivities()
+                ActivityInfoTree.filterEnabledActivities()
+            }
```
petitlapin: Here you should call ActivityInfoTree.filterBySearch(input.text) in the else case. Because if the user changes the conf while in the search, the changes won't be reflected

src/activities/menu/Menu.qml
```diff
@@ -459,6 +500,136 @@ ActivityBase {
 
         }
 
+
+        Rectangle
```
petitlapin: put the accolade in the same line as the item: Rectangle {
Same for below with the TextInput and VirtualKeyboard

src/core/ActivityInfoTree.cpp
```diff
@@ -283,5 +285,71 @@ void ActivityInfoTree::init()
 
     qmlRegisterSingletonType<QObject>("GCompris", 1, 0, "ActivityInfoTree", menuTreeProvider);
     qmlRegisterType<ActivityInfo>("GCompris", 1, 0, "ActivityInfo");
+
+
+}
+void ActivityInfoTree::filterBySearch(const QString& text)
+{
+
+    if(text.trimmed().length() < m_searchedtext.length() or text.trimmed() == ""){
```
petitlapin: If I remember, we said it was not useful to optimize?
It causes also issues if when you are in a search: let's say you typed "chess", you select the 'e' and type 'a'. No activities should match but in your case, nothing will change

src/core/ActivityInfoTree.cpp
```diff
@@ -283,5 +285,71 @@ void ActivityInfoTree::init()
...
+    m_searched = m_menuTree;
+    filterEnabledActivities();
+    filterLockedActivities();
+    sortByDifficulty();
+    emit menuTreeChanged();
+
+}
+QVariantList  ActivityInfoTree::allCharacters()
+{
+    int i = 0;
```
petitlapin: unused?

src/core/ActivityInfoTree.cpp
```diff
@@ -283,5 +285,71 @@ void ActivityInfoTree::init()
...
+    QSet<QString> keyboardchars;
+    QString templetter;
+    foreach(auto tree,m_menuTreeFull)
+    {
+        auto title = tree->title();
+        for (int j = 0 ; j!= title.length(); j++)
+        {
+            if(!title.at(j).isSpace() and !title.at(j).isPunct())
+            {
+                templetter = title.at(j).toLower();
```
petitlapin: you can directly insert title.at(j).toLower() without having to use a temporary variable

src/core/ActivityInfoTree.h
```diff
@@ -31,11 +31,12 @@ class ActivityInfoTree : public QObject
 	Q_PROPERTY(ActivityInfo* rootMenu READ getRootMenu CONSTANT)
 	Q_PROPERTY(QQmlListProperty<ActivityInfo> menuTree READ menuTree NOTIFY menuTreeChanged)
     Q_PROPERTY(ActivityInfo* currentActivity READ getCurrentActivity WRITE setCurrentActivity NOTIFY currentActivityChanged)
+    Q_PROPERTY(QVariantList characters READ allCharacters)
```
petitlapin: the names are not explicit enough

src/core/ActivityInfoTree.h
```diff
@@ -46,6 +47,8 @@ class ActivityInfoTree : public QObject
 						const QDir &menuDir, const QString &menuFile);
 	void sortByDifficulty();
 	void sortByName();
+    QVariantList allCharacters();
+    QVariantList keyboardcharacters;
```
petitlapin: Need to be prefixed with m_ to know it's a class attribute. Need to use camelCase also for the name.

src/core/ActivityInfoTree.h
```diff
@@ -61,12 +65,15 @@ protected slots:
 private:
     // this is the full activity list, it never changes
     QList<ActivityInfo *> m_menuTreeFull;
+    QList<ActivityInfo *> m_menuTreeFulltemp;
```
petitlapin: if it was for optimization, we can remove it along with m_searched and m_searchedtext

src/core/VirtualKeyboard.qml
```diff
@@ -326,18 +333,18 @@ Item {
          */
     }
 
-    Behavior on height {
-        NumberAnimation {
-            duration: 500
-            easing.type: Easing.OutCubic
-        }
-    }
+//    Behavior on height {
```
petitlapin: why comment these? (I don't know what was the use)

src/core/ActivityInfoTree.cpp
rish9511: the current code will not display any activity if you delete 'e' and type 'a' but  if you delete 
'a' and type 'e' again in this case no activity will be displayed although all the activities with the name 'chess' should have been displayed .So yes there is issue with the current code but I have come up with a solution to handle such cases so if it's okay we can keep doing the optimization with handling the case that you  have mentioned.

Reason why I would like to keep the optimized code:

1. If the user is using virtual keyboard there is no way for him to go back and delete a specific character   that's because we disable the focus if user has opted to use  virtual  keyboard and if there is no focus the cursor won't be visible .

2. ex: if I type "planefame"  instead of "planegame"  I would delete 'fame' not just 'f' and type 'game'    ,this I what I usually do  unless I  typed a very long string and made a mistake of mistyping a single character but we can't expect every user to do that .Many would prefer deleting that single mistyped character and my new approach handles this case :)

src/activities/menu/Menu.qml
rish9511: does m_filterLevelMin changes on changing the difficulty level from configuration?

src/core/ActivityInfoTree.h
rish9511: I am bad at this :P any suggestions?

src/core/ActivityInfoTree.h
rish9511: is the name fine ?

src/activities/menu/Menu.qml
```diff
@@ -77,6 +78,8 @@ ActivityBase {
 
     // @cond INTERNAL_DOCS
     property string url: "qrc:/gcompris/src/activities/menu/resource/"
+    property int  cursorPosition: 0
+    property  string inputText: ""
```
petitlapin: You can set it as an alias of the textInput.text (or the opposite), this way, it will be automatically updated.

src/core/ActivityInfoTree.cpp
```diff
@@ -139,16 +139,35 @@ void ActivityInfoTree::filterByTag(const QString &tag)
     for(auto activity: m_menuTreeFull) {
         if((activity->section().indexOf(tag) != -1 ||
 			tag == "all" ||
-            (tag == "favorite" && activity->favorite())) &&
-            (activity->difficulty() >= ApplicationSettings::getInstance()->filterLevelMin() &&
-             activity->difficulty() <= ApplicationSettings::getInstance()->filterLevelMax())) {
-            m_menuTree.push_back(activity);
+            (tag == "favorite" && activity->favorite()))) {
+            filterByDifficulty(activity,ApplicationSettings::getInstance()->filterLevelMin(),
```
petitlapin: I wouldn't filter by difficulty activity by activity but only once at the end of the for loop

src/core/ActivityInfoTree.cpp
```diff
@@ -283,5 +302,70 @@ void ActivityInfoTree::init()
 
     qmlRegisterSingletonType<QObject>("GCompris", 1, 0, "ActivityInfoTree", menuTreeProvider);
     qmlRegisterType<ActivityInfo>("GCompris", 1, 0, "ActivityInfo");
+
+
+}
+void ActivityInfoTree::filterBySearch(const QString& text,int cursorPosition)
```
petitlapin: the cursor position shouldn't be a condition on the filtering. There is no "logical" reason to have it here. (I understand you use it for the if-case but it does not fit here)

src/core/ActivityInfoTree.cpp
```diff
@@ -283,5 +302,70 @@ void ActivityInfoTree::init()
...
+    emit menuTreeChanged();
+
+}
+QVariantList  ActivityInfoTree::allCharacters()
+{
+    QSet<QString> keyboardchars;
+    foreach(auto tree,m_menuTreeFull)
+    {
+        auto title = tree->title();
+        for (int j = 0 ; j!= title.length(); j++)
```
petitlapin: not necessary for now I think, but you probably can make it more "c++11" :)

bdoin:
I tested it. My remarks:
- the search is done on title and name. It should search also on goal, manual and section.
- the keyboard is not working
- if we enter 2 words we expect to filter more. You must run the filter on each words in searchedText.

src/core/ActivityInfoTree.cpp
rish9511: I'll make this change

src/core/ActivityInfoTree.cpp
rish9511: any other way through which we can access cursor position in activityinfotree.cpp? Using q_property ?

src/activities/menu/Menu.qml
rish9511: on setting inputText to input.text I get this error
"ReferenceError: input is not defined"

src/activities/menu/Menu.qml
petitlapin: yes

src/core/ActivityInfoTree.cpp
```diff
@@ -283,5 +291,53 @@ void ActivityInfoTree::init()
...
+                    activity->description().contains(text.trimmed(),Qt::CaseInsensitive)){
+
+                m_searchedActivities.push_back(activity);
+            }
+        }
+
+    }
+    else
+        m_searchedActivities = m_menuTreeFull;
+    m_menuTree = m_searchedActivities;
```
petitlapin: do you need to have m_searchedActivities tree?

src/activities/menu/Menu.qml
petitlapin: yes, because it is embedded. Then, you can set the TextInput.text value to inputText (maybe rename it searchedText?), so you won't have to update it manually. The aim is to have only one variable to update

src/activities/menu/Menu.qml
```diff
@@ -459,6 +500,131 @@ ActivityBase {
...
+               GradientStop { position: 1.0; color: "#AAA" }
+
+           }
+
+            TextField{
+                id : input
+                anchors.fill: parent
+                text: ""
+                textColor: "black"
+                font.pixelSize: 28
```
bdoin: You must not use pixelSize, it does not scale properly on HDPI devices. Use fontSize instead like this:
font.pointSize: (18 + ApplicationSettings.baseFontSize) * ApplicationInfo.fontRatio

src/core/ActivityInfoTree.cpp
```diff
@@ -283,5 +291,56 @@ void ActivityInfoTree::init()
 
     qmlRegisterSingletonType<QObject>("GCompris", 1, 0, "ActivityInfoTree", menuTreeProvider);
     qmlRegisterType<ActivityInfo>("GCompris", 1, 0, "ActivityInfo");
+
+
+}
+void ActivityInfoTree::filterBySearch(const QString& text)
+{
+    m_menuTree.clear();
+    if(!text.trimmed().isEmpty()){
```
bdoin: Our formatting rules requires a space before a '{' and a space after a ','.

bdoin:
In your commit name you must put your first and last name. This is mandatory for KDE.
When we enter the search menu for the first time the focus is not set on the search entry.
The allCharacters function seems not to be use. I tested with the virtual keyboard and not letters appears in the keyboard.

PR #112, opened 2016-03-01
demo guesscount activity

rahulyadav170923:
demo guesscount activity


bdoin:
I tried it, there are many error in the console:
qrc:/gcompris/src/activities/guesscount/Guesscount.qml:196:86: Unable to assign QQuickItem to QQuickAnchorLine
qrc:/gcompris/src/activities/guesscount/Guesscount.qml:54: TypeError: Cannot read property '0' of undefined
qrc:/gcompris/src/activities/guesscount/Guesscount.qml:55: TypeError: Cannot read property '1' of undefined
qrc:/gcompris/src/activities/guesscount/Guesscount.qml:45:9: QML QtObject: Binding loop detected for property "question_no"

This is a good start. What's missing is that the levels should be calculated and at upper levels there should be multiple lines of questions.

PR #115, opened 2016-03-04
Rahulyadav

rahulyadav170923:
sir the activity is fully functional. but if take a look at Application output there is one single error repeated many times(QML Rectangle: Cannot anchor to a null item.).i have not been successfull in solving this error. please let me know if i should make any changes in the working of the activity. i am starting to work on sizing it properly and making a blueprint for practice mode. i,ll let you know the details soon.


petitlapin:
Hi,

It's a nice activity, which does what it's supposed to do.

a few remarks to improve it after playing with it:

I know the backgrounds are the Gtk+ one, but I don't think they are adapted to this activity, maybe we can have something more like a black board or a tablet (just a proposition, don't hesitate to say if you have other ideas)?
There should be some protection when displaying some results: for example, 8 : 3 will display 2.6666666666666665 which is not good. I'm not sure if the best solution would be to display a fraction or an approximation, or even not consider non integer results for division?
We should be able to replace an operator by drag and dropping an other one on an existing one (I mean if you have "3" "*" "2", you should be able to drop "+" directly at the place of "*".
If I have two computation lines, if I drag one element in one of the 3 possible places, sometimes, I can't move to some cases. I think it's because the first line is not fully filled but it's not intuitive. And sometimes the result is not computed. I think it may be related to the "Cannot anchor to a null item". I'll need to dig in the code to see.
At latest sublevel of level 3, if I only put 10-3 at first line I win.

I didn't have time to closely look at the code, but in qml, you can probably use a Repeater for the computation rows and there are too many harcoded sizes ;).
In js file, I like your shuffle function, but we already have one in Core that really shuffles :D.

Johnny

rahulyadav170923:
thanks for the review johnny. i don't know how i left (10-3) case. i make the changes and send a pull request by the end of the day.

src/activities/activities.txt
```diff
@@ -51,6 +51,7 @@ gnumch-inequality
 gnumch-multiples
 gnumch-primes
 graph-coloring
+guesscount
```
petitlapin: you forgot to add the practise activity in this file

src/activities/guesscount/ActivityInfo.qml
```diff
@@ -0,0 +1,35 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+import GCompris 1.0
+
+ActivityInfo {
+    name: "guesscount/Guesscount.qml"
+    difficulty: 1
+    icon: "guesscount/guesscount.svg"
+    author: "Your Name &lt;yy@zz.org&gt;"
```
petitlapin: update your name here

src/activities/guesscount/ActivityInfo.qml
```diff
@@ -0,0 +1,35 @@
...
+    demo: true
+    title: qsTr("Guessing the algebraic expression for the answer")
+    description:qsTr("guess the algebraic expression and click on the buttons to add them.")
+    //intro: "guesscount activity"
+    goal: qsTr("To teach algebra")
+    prerequisite: qsTr("knowledge of arithematic oprations")
+    manual: ""
+    credit: ""
+    section: "math"
+    createdInVersion:1
```
petitlapin: this should be 6000 (it's the GCompris version number where the activity is created, not an activity versioning)

src/activities/guesscount/Guesscount.qml
```diff
@@ -0,0 +1,973 @@
...
+            spacing: 460
+            anchors{
+                top:parent.top
+                topMargin:30
+                left:parent.left
+                leftMargin: 34
+            }
+            Rectangle {
+                id:question_no
+                width:400;
```
petitlapin: this won't work when resizing the screen

src/activities/guesscount/Guesscount.qml
```diff
@@ -0,0 +1,973 @@
...
+                                ParentChange { target: rec; parent: root }
+                                AnchorChanges { target: rec; anchors.verticalCenter: root.anchors.verticalCenter; anchors.horizontalCenter: root.anchors.horizontalCenter }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        // row 3
+        Row {
```
petitlapin: you should probably also use a Repeater for the Row

src/activities/guesscount/Guesscount.qml
```diff
@@ -0,0 +1,973 @@
...
+
+                            }
+                            else if(items.no_of_rows==3)
+                            {
+                                audioEffects.play("qrc:/gcompris/src/core/resource/sounds/crash.wav")
+                            }
+
+
+                        }
+
```
petitlapin: For all the file, don't put a lot extra line breaks.

src/activities/practice/practice.js.autosave
```diff
@@ -0,0 +1,76 @@
+/* GCompris - practice.js
```
petitlapin: don't commit .autosave files

src/activities/practice/practice.js
```diff
@@ -0,0 +1,77 @@
+/* GCompris - practice.js
+ *
+ * Copyright (C) 2016 YOUR NAME <xx@yy.org>
```
petitlapin: your name :)

src/activities/practice/Practice.qml.autosave
```diff
@@ -0,0 +1,515 @@
+/* GCompris - practice.qml
```
petitlapin: don't commit .autosave files

src/activities/practice/Practice.qml
```diff
@@ -0,0 +1,692 @@
+/* GCompris - practice.qml
+ *
+ * Copyright (C) 2016 YOUR NAME <xx@yy.org>
```
petitlapin: your name :)

src/activities/practice/Practice.qml
```diff
@@ -0,0 +1,692 @@
...
+                    height: 40
+                    signal clicked
+                    onClicked: NumberAnimation { target: evaluate; property: "opacity"; from: 0; to: 1; duration: 200 }
+
+                    color: "#53d769"
+                    border.color: Qt.lighter(color, 1.1)
+
+                    GCText {
+                        anchors.centerIn: parent
+                        text: "Evaluate"
```
petitlapin: qsTr() missing

src/activities/practice/Practice.qml
```diff
@@ -0,0 +1,692 @@
...
+                    height: 50
+                    signal clicked
+                    onClicked: NumberAnimation { target: correctanswers; property: "opacity"; from: 0; to: 1; duration: 200 }
+
+                    color: "#53d769"
+                    border.color: Qt.lighter(color, 1.1)
+
+                    GCText {
+                        anchors.centerIn: parent
+                        text: "Correct Answers"
```
petitlapin: missing qsTr()

src/activities/practice/Practice.qml
```diff
@@ -0,0 +1,692 @@
...
+                    }
+                    MouseArea {
+                        anchors.fill: parent
+                        hoverEnabled: true
+                        onHoveredChanged: {
```
petitlapin: I would do the opposite:
in correctanswers, I would do, scale: mouseArea.containsMouse ? 1.15 : 1

src/activities/practice/ActivityInfo.qml
```diff
@@ -0,0 +1,35 @@
...
+ */
+import GCompris 1.0
+
+ActivityInfo {
+  name: "practice/Practice.qml"
+  difficulty: 1
+  icon: "practice/practice.svg"
+  author: "Your Name &lt;yy@zz.org&gt;"
+  demo: true
+  title: "Practice activity"
```
petitlapin: qsTr()

src/activities/practice/ActivityInfo.qml
```diff
@@ -0,0 +1,35 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+import GCompris 1.0
+
+ActivityInfo {
+  name: "practice/Practice.qml"
+  difficulty: 1
+  icon: "practice/practice.svg"
+  author: "Your Name &lt;yy@zz.org&gt;"
```
petitlapin: your name

src/activities/practice/ActivityInfo.qml
```diff
@@ -0,0 +1,35 @@
...
+  demo: true
+  title: "Practice activity"
+  description: qsTr("add the numbers and operators yourself and check if your answer is correct")
+  //intro: "put here in comment the text for the intro voice"
+  goal: qsTr("to build confidence in children while solving arithematic operations")
+  prerequisite: qsTr("arithematic operations")
+  manual: ""
+  credit: ""
+  section: "math"
+  createdInVersion:1
```
petitlapin: 6000

petitlapin:
Hi,

you can remove the tiger images if you don't use them anymore.

For the practice activity, I have some doubts:
If you do it as an activity, it should be in another PR, because this one should only contains code related to guesscount.
And even as a mode, I'm not sure of the usefulness like that. It's a calculator where the child has to put the answer he wants, because he has no feedback if the answer is wrong (what is the good answer).
If I type 66 + 66 = 132 and press evaluate, the second number displays 6 instead of being empty.
If I type 99 * 99, I can't type more than 99 in the answer field (seems limited to 200).

You should take care of the activities resizing (both ;)). For guesscount, it does not resize at all and for practice everything gets mixed.

Johnny

rahulyadav170923:
thanks johnny for the review i'll make the changes as you suggested. about
the practice activity i'll add option to see correct answer and work on
screen compatibility. it might take some time as my mid semester exams have
started(13-19). and i have put a validation on numbers(but i'll ask about
it to bdoin also and then make changes).
thnx

PR #123, opened 2016-03-12
Jumbled words activity

KomalParmaar:
activity under reading section.
only English language right now.


src/activities/jumbled_words/ActivityInfo.qml
```diff
@@ -0,0 +1,36 @@
...
+ */
+import GCompris 1.0
+
+ActivityInfo {
+  name: "jumbled_words/Jumbled_words.qml"
+  difficulty: 1
+  icon: "jumbled_words/jumbled_words.svg"
+  author: "Komal Parmaar &lt;parmaark@gmail.com&gt;"
+  demo: true
+  title: "Jumbled letters"
```
petitlapin: use qsTr() for translatable strings

src/activities/jumbled_words/ActivityInfo.qml
```diff
@@ -0,0 +1,36 @@
...
+  description: "Type in the correct meaningful word from the given jumbled word.
+  A hint box is provided below the jumbled word. Click on the hint box to get a hint."
+  //intro: "put here in comment the text for the intro voice"
+  goal: "To identify all the jumbled words correctly"
+  prerequisite: "None"
+  manual: ""
+  credit: ""
+  section: "reading"
+  createdInVersion: 1000
```
petitlapin: you should not have changed this attribute, you should leave 6000

src/activities/jumbled_words/Jumbled_words.qml
```diff
@@ -0,0 +1,473 @@
+/* GCompris - jumbled_words.qml
+ *
+ * Copyright (C) 2016  Komal Parmaar <parmaark@gmail.com>
+ *
+ * Authors:
+ *   <THE GTK VERSION AUTHOR> (GTK+ version)
```
petitlapin: remove this line as this is not a port

src/activities/jumbled_words/Jumbled_words.qml
```diff
@@ -0,0 +1,473 @@
...
+                opacity: 0.31
+                radius: 10
+
+                MouseArea {
+                    id: mouseArea
+                    anchors.fill: parent
+                    hoverEnabled: ApplicationInfo.isMobile ? false : true
+
+                    onClicked: {
+                        if (Activity.checkAnswer()) {
```
petitlapin: why an empty if?

src/activities/jumbled_words/Jumbled_words.qml
```diff
@@ -0,0 +1,473 @@
...
+                }
+            }
+
+            GCText {
+                id: questionText
+                anchors.horizontalCenter: parent.horizontalCenter
+                anchors.verticalCenter: parent.verticalCenter
+                opacity: 1.0
+                z:11
+                text: ""
```
petitlapin: no need to initialize it to ""

src/activities/jumbled_words/Jumbled_words.qml
```diff
@@ -0,0 +1,473 @@
...
+                }
+            }
+
+            GCText {
+                id: hintText
+                anchors {
+                    horizontalCenter: parent.horizontalCenter
+                }
+                opacity: 1.0
+                text: "hint"
```
petitlapin: qsTr() for translatable strings

src/activities/jumbled_words/jumbled_words.js
```diff
@@ -0,0 +1,185 @@
...
+    loadLevels();
+    currentLevel = 0;
+    currentSubLevel = 0;
+    maxLevel = levels.length;
+    initLevel()
+}
+
+function loadLevels()
+{
+    var filename = GCompris.ApplicationInfo.getLocaleFilePath(url + "levels-$LOCALE.json");
```
petitlapin: why don't you directly use the datasets of wordsgame or lang?

src/activities/jumbled_words/jumbled_words.js
```diff
@@ -0,0 +1,185 @@
...
+            level = levels[currentLevel];
+            maxSubLevel = level.questions.length;
+            items.score.numberOfSubLevels = maxSubLevel;
+            items.score.currentSubLevel = "1";
+            answers = level.questions[Math.floor(Math.random()*maxSubLevel)];
+            var copy = answers;
+        var demo = shuffleString(answers);
+        if(demo == answers)
+        {
+            demo = shuffleString(answers);
```
petitlapin: what happens if it is still the same? a do while{} loop seems better?

src/activities/jumbled_words/jumbled_words.js
```diff
@@ -0,0 +1,185 @@
...
+    }
+    return a.join("");
+}
+
+function initLevel() {
+    items.bar.level = currentLevel + 1
+    items.hint.text = "hint";
+
+    if (currentSubLevel == 0) {
+            level = levels[currentLevel];
```
petitlapin: keep the same indentation everywhere for this function

src/activities/jumbled_words/jumbled_words.js
```diff
@@ -0,0 +1,185 @@
...
+}
+
+function checkAnswer()
+{
+    var modelEntry = items.edit.getText(0, items.edit.text.length)
+    var value = modelEntry.replace(/\s\s*$/, '') //to remove leading white spaces
+    value = modelEntry.replace(/^\s\s*/, '') //to remove trailing white spaces
+    if (value === answers) {
+        items.bonus.good("flower");
+        items.edit.text = "";
```
petitlapin: you should avoid setting variables in js because this breaks the qml binding. Can you do it in qml with a condition?
something like, GCText { id: edit; text: condition?qsTr("text1") : "" ... }

src/activities/jumbled_words/ActivityInfo.qml
```diff
@@ -0,0 +1,36 @@
...
+import GCompris 1.0
+
+ActivityInfo {
+  name: "jumbled_words/Jumbled_words.qml"
+  difficulty: 1
+  icon: "jumbled_words/jumbled_words.svg"
+  author: "Komal Parmaar &lt;parmaark@gmail.com&gt;"
+  demo: true
+  title: "Jumbled letters"
+  description: "Type in the correct meaningful word from the given jumbled word.
```
petitlapin: you forgot the closing "

petitlapin:
Hi,

a few more comments after testing it:
* you need to have a DialogHelp in your main qml. Else, when you click on the Help button, it does display the general help and not the activity one.
* I have these warnings when starting: "QQuickWorkerScript: Attempt to send message before WorkerScript establishment" and "qrc:/gcompris/src/activities/jumbled_words/jumbled_words.js:123: Error: Cannot assign [undefined] to QString" (which should be fixed using the do while loop)
* hjg.svg is unused, you should remove it.
* I don't understand the game aim, is the game to type the word given? Sometimes, it does not work (I have "animal" in the cloud, I type it and nothing happen).
I had at first "hint" to type, I typed it and it shows me "animal" now
-> after a few minutes, I understand, the shuffled letters are not displayed at first time (probably because of the warning).
* the interface is too minimalistic. You can probably do the same as for hangman (with an image as hint). The text in the input panel should have the same size as the jumbled text. You can also probably add a background to the input panel.
* sometimes, there are multiple times the same word to find in the same level. They should all be different.
* there should be the configuration button where we can choose another language.
* the virtual keyboard is not displayed even if I have the settings active in the configuration.
* There should be an OK button where the child has to click to validate the word he/she has found.

Johnny

KomalParmaar:
Please review the update commit. 
I have tried to update all the points.

src/activities/jumbled_words/ActivityInfo.qml
```diff
@@ -0,0 +1,37 @@
...
+ * along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+import GCompris 1.0
+
+ActivityInfo {
+  name: "jumbled_words/Jumbled_words.qml"
+  difficulty: 1
+  icon: "jumbled_words/jumbled_words.svg"
+  author: "Komal Parmaar &lt;parmaark@gmail.com&gt;"
+  demo: false
```
petitlapin: This should be set to true. This property tells if the activity is accessible by default on the gratis version or not.

src/activities/jumbled_words/Jumbled_words.qml
```diff
@@ -0,0 +1,486 @@
...
+                function insertText(text) {
+                    edit.insert(cursorPosition, text)
+                }
+                function backspace() {
+                    if(cursorPosition > 0) {
+                        moveCursorSelection(cursorPosition - 1, TextEdit.SelectCharacters)
+                        cut()
+                    }
+                }
+                function formatLineWith(tag) {
```
petitlapin: do you need/use this function?

src/activities/jumbled_words/Jumbled_words.qml
```diff
@@ -0,0 +1,486 @@
...
+            radius: 10
+
+            MouseArea {
+                id: mouseAreaSubmit
+                anchors.fill: parent
+                hoverEnabled: ApplicationInfo.isMobile ? false : true
+                onClicked: {
+                    if(Activity.checkAnswer())
+                    {
+                        edit.text=""
```
petitlapin: you should not set the values here but probably use a boolean hintDisplayed and the hintText.text should be something like: hintDisplayed ? theHint : qsTr("Hint")

src/activities/jumbled_words/Jumbled_words.qml
```diff
@@ -0,0 +1,486 @@
...
+                        hintText.text="Hint"
```
petitlapin: you should not set the values here but probably use a boolean hintDisplayed and the hintText.text should be something like: hintDisplayed ? theHint : qsTr("Hint")

src/activities/jumbled_words/Jumbled_words.qml
```diff
@@ -0,0 +1,486 @@
...
+                displayDialog(dialogHelpLeftRight)
+            }
+            onHomeClicked: activity.home()
+            onReloadClicked: {
+               edit.text = ''
+               hintText.text = 'Hint';
+               hintImage.source="";
+            }
+            onPreviousLevelClicked: {
+                hintImage.source="";
```
petitlapin: you can probably do this in the initLevel function instead of doing it in each button

src/activities/jumbled_words/jumbled_words.js
```diff
@@ -0,0 +1,177 @@
...
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+.pragma library
+.import QtQuick 2.0 as Quick
+.import GCompris 1.0 as GCompris
+.import QtGraphicalEffects 1.0 as QtGraphicalEffects
```
petitlapin: I'm not sure if this import is required?

src/activities/jumbled_words/jumbled_words.js
```diff
@@ -0,0 +1,177 @@
...
+var defaultLevelsFile = "qrc:/gcompris/src/activities/lang/resource/words_sample.json";
+
+var currentLevel
+var maxLevel
+var maxSubLevel
+var currentSubLevel
+var items
+var levels;
+var level;
+var temp;
```
petitlapin: Not a good name for a variable.
For all the variables, are you sure they need to be global?

src/activities/jumbled_words/jumbled_words.js
```diff
@@ -0,0 +1,177 @@
...
+    if (currentSubLevel == 0) {
+        level = levels[currentLevel];
+        maxSubLevel = level.content[0].content.length;
+        for(var i=0;i< maxSubLevel ; i++)
+        {
+            array[i]=i;
+        }
+        array=Core.shuffle(array);
+        items.score.numberOfSubLevels = maxSubLevel;
+        items.score.currentSubLevel = "1";
```
petitlapin: it's an integer, not a string

src/activities/jumbled_words/jumbled_words.js
```diff
@@ -0,0 +1,177 @@
...
+        temp=array[currentSubLevel];
+        answers = level.content[0].content[temp].description;
+        copy = answers;
+        var demo = shuffleString(answers);
+        if(demo == answers)
+        {
+            demo = shuffleString(answers);
```
petitlapin: what happens if the demo is still the same as answers after the second shuffle?

src/activities/jumbled_words/jumbled_words.js
```diff
@@ -0,0 +1,177 @@
...
+        }
+        else
+            questions = demo;
+    }
+    else {
+        do{
+            temp=temp=array[currentSubLevel];
```
petitlapin: 2 assignments?

src/activities/jumbled_words/jumbled_words.js
```diff
@@ -0,0 +1,177 @@
...
+            generate = level.content[0].content[temp].description;
+
+        } while(copy===generate);
+        copy=generate;
```
petitlapin: what is the use of copy, generate...? You seem to have a lot of variables doing more or less the same.

src/activities/jumbled_words/jumbled_words.js
```diff
@@ -0,0 +1,177 @@
...
+        nextLevel()
+    }
+    initLevel();
+}
+
+function checkAnswer()
+{
+    var modelEntry = items.edit.getText(0, items.edit.text.length)
+    var value = modelEntry.replace(/\s\s*$/, '') //to remove leading white spaces
+    value = modelEntry.replace(/^\s\s*/, '') //to remove trailing white spaces
```
petitlapin: shouldn't it be value.replace(/^\s\s*/, '')? else the first replace isn't used

KomalParmaar: yes, right!  note taken.

src/activities/jumbled_words/Jumbled_words.qml
KomalParmaar: Will it not break the qml binding which you had mentioned on the last commit?

src/activities/jumbled_words/Jumbled_words.qml
KomalParmaar: actually i realized there is no need to set the value to hinText.text to hint again since the text does not change on clicking, just the image is displayed on one corner of screen.

src/activities/jumbled_words/Jumbled_words.qml
KomalParmaar: No. Removed.

petitlapin:
Hi,

* When resizing I don't see the word I'm typing anymore.
* Return should validate the word like a click on Submit.
*Virtual keyboard should be displayed when the option is activated in configuration.
* When inputting the good word and clicking on the jumbled text, the image and the "good" flower appear but the text is still there for the new level, it should be erased.
* The text in the input panel should be centered, not draggable and if possible limited to the number of letters that exist. Potential improvement: colorize in the upper text the letters inputted in the bottom and only accept the letter of the word to find? What do you think of it?
* I don't understand why the input panel is wider than the provided text.
* It is purely personal but I don't like the layout of the activity. I think we need more advice for it, what do you think?
* I have this error when changing level qrc:/gcompris/src/activities/jumbled_words/jumbled_words.js:104: TypeError: Cannot read property 'description' of undefined
* After changing language in parameter, I have qrc:/gcompris/src/activities/jumbled_words/jumbled_words.js:156: TypeError: Cannot read property 'image' of undefined when I click on submit
* Sometimes I have twice the same word in the same session.
* The language change does not work (I have French by default but words are in english)

Johnny

src/activities/jumbled_words/ActivityInfo.qml
```diff
@@ -0,0 +1,37 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+import GCompris 1.0
+
+ActivityInfo {
+  name: "jumbled_words/Jumbled_words.qml"
+  difficulty: 1
```
petitlapin: I would update difficulty because 1 is when children don't need to know reading

src/activities/jumbled_words/Jumbled_words.qml
```diff
@@ -0,0 +1,403 @@
...
+
+        Bar {
+            id: bar
+            content: BarEnumContent { value: help | home | level | reload | config}
+            onHelpClicked: {
+                displayDialog(dialogHelpLeftRight)
+            }
+            onHomeClicked: activity.home()
+            onReloadClicked: {
+               edit.text = Qt.binding(function() { return ""})
```
petitlapin: binding are supposed to be used with variables that can change not with static value. In this case, it is not useful, you can remove them

KomalParmaar:
Yes, the color change on each correct word is a great idea.
Will do the changes by this weekend for sure.
Thanks for reviewing.

bdoin:
Hi, tested you activity. It start to be interesting. You should consider making it easier for children. For example starting with less jumble in the words. You could also do like in the hangman and make the image to find  appear smoothly. Another idea is to present several words. This would let the children read other words and find the closer ones. You could put the images as well as the other words.

PR #124, opened 2016-03-14
Multiplication tables

varun13169:
Hi,
I build this activity based on JohnnyJ's suggestion: http://www.infomontessori.com/mathematics/tables-of-arithmetic-multiplication-board.htm

In this a child has to count and mark squares in order to match the answer given in the table.


petitlapin:
Hi,
just a quick note but be careful at your commit messages, they should describe more what you do than your emotional state ;)

src/activities/align4-2players/Piece.qml
```diff
@@ -35,7 +35,7 @@ Image {
             name: "invisible"
             PropertyChanges {
                 target: piece
-                opacity: 0
+                opacity: 1
```
petitlapin: why do you have changes in other activities?

src/activities/multiplication-tables/MultiplicationTables.qml
```diff
@@ -0,0 +1,254 @@
...
+                            name: "active"
+                            PropertyChanges { target: dots; color: "red"}
+                            PropertyChanges { target: dots; clickedFlag: true}
+                        }
+                    ]
+
+                    MouseArea {
+                        anchors.fill: parent
+                        onClicked: {
+                            if (dots.state == "default")
```
petitlapin: If I understand your code (sorry, didn't have time to test it yet), you need to press each case to get the good result (which will be quite boring to higher levels)?
I think we can do it more interactive as in the link: the child has to place 2 cards on top and left, we colorize automatically the beams between and the child has to input the result. What do you think?

varun13169: From what I understood:
According to Montessori, aim for this activity is to count the number of squares(beads in Montessori's site) in each case, and table is explicitly given to the child for help.
Now, as he counts the number of squares by physically clicking them, he can telly his answer with the table given on the right(Point number 8 on Montessori's site), thereby relating the concept of tables with the multiplication and addition.

On the boring part, even I felt it monotonous to just punch and count squares from level 1-10, but for a specific table I think it's a good way to visualize how addition of a specific number multiple times is related to the table of that number.
**Solution:** Instead of relating tables with levels, we let the child decide which table he/she wants, just like sliding the card(Point number 3 on Montessori's site).

On making it more interactive, in which child enters the solution.
**Solution:** If tables are not related to levels, as suggested in above solution, then I can try making two levels, one in which tables are shown(just like the present game) and second in which along with counting squares by clicking them, a child needs to input solution also, eg:- instead of "4 x 5 = 20", "4 x 5 = " is given.

Please let me know your thoughts on the this.

src/activities/align4-2players/Piece.qml
varun13169: Reverted all the changes.
Actually I was experimenting with this game, as I changed everything back to what it was, hence didn’t feel the need to check the difference, will take care of this from next time.

bdoin:
I play with it and I don't like the game play. It implies way too many clicks. A drag move would be better. But I don't understand very well the goal of the game. At some point when the children find the right spots for 3x4 it should at least be asked for the answer?

PR #125, opened 2016-03-21
Multiplication Chart

varun13169:
This activity is based on Montessori Multiplication Chart Activity:
http://www.infomontessori.com/mathematics/tables-of-arithmetic-multiplication-charts.htm

A child will be given a multiplicand and a multiplier, he/she needs to select row and column corresponding to either multiplicand or multiplier and then report back the answer.

Here is how it looks.
layout


petitlapin:
Hi,

you can probably center the texts in the cases, it should be better.

Johnny

src/activities/multiplication-chart/MultiplicationChart.qml
```diff
@@ -0,0 +1,287 @@
...
+ActivityBase {
+    id: activity
+
+    onStart: focus = true
+    onStop: {}
+
+    pageComponent: Image {
+        id: background
+        fillMode: Image.PreserveAspectCrop
+        source: "./resource/numbers.svg"
```
petitlapin: you should use the qrc path ("qrc:/gcompris/src/activities/{activity}/resource/").

src/activities/multiplication-chart/MultiplicationChart.qml
```diff
@@ -0,0 +1,287 @@
...
+            anchors {
+                top: parent.top
+                topMargin: 20
+                left: parent.left
+                leftMargin: 20
+            }
+            spacing: items.spacing
+            columns: 1
+            rows: 11
+            Repeater {
```
petitlapin: As it is the same code as above, you probably can do a separate item for it

src/activities/multiplication-chart/multiplication-chart.js
```diff
@@ -0,0 +1,129 @@
...
+
+function stop() {
+}
+
+function initLevel() {
+    items.bar.level = currentLevel + 1
+    items.rowSelected = 0
+    items.colSelected = 0
+    items.enteredInput = 0
+    items.answer.text = qsTr("")
```
petitlapin: no need of qsTr for empty strings

src/activities/multiplication-chart/multiplication-chart.js
```diff
@@ -0,0 +1,129 @@
...
+        currentLevel = numberOfLevel - 1
+    }
+
+    initLevel();
+}
+
+
+
+function makeOtherColInRowWhite() {
+    for(var i=0; i<items.repeaterGridRow.count; i++) {
```
petitlapin: Can't the state be directly binded to the value of the selected row/column? I mean, in the Repeater having something like:
state: index == items.rowSelected ? "active" : "default"

src/activities/multiplication-chart/MultiplicationChart.qml
```diff
@@ -0,0 +1,287 @@
...
+                horizontalCenter: instruction.horizontalCenter
+                horizontalCenterOffset: -30
+                top: instruction.bottom
+                topMargin: 40
+            }
+            fontSize: hugeSize
+            text: items.rowQues + " X " + items.colQues + " = "
+        }
+
+        TextInput {
```
petitlapin: We have a Numpad item in core that could be better than having a TextInput, what do you think?

varun13169:
Thanks for the suggestions.
Will look at them and will incorporate them in a week :)

varun13169:
Hi,
1) I added numpad to this activity as was suggested, now it function similar to algebra_by.

2) Moved the answer in middle, changed instruction font to small.

3) I tried to remove the loop in JS that was changing the state but it didn’t work well, reason is when mouse is clicked index have the value of the current rectangle only, so no knowledge about past is there hence loop is needed.
If I keep it outside the mouse area then it works but only for first iteration at the time of initialisation but I need it to check whenever mouse is clicked.

Now activity looks like this.
screenshot from 2016-03-28 00 45 15

PR #127, opened 2016-03-27
Multiplication chart

varun13169:
Hi,

1) I added numpad to this activity as was suggested, now it function similar to algebra_by.

2) Centred the texts in the cases.

3) Moved the answer in middle, changed instruction font to small.

4) I tried to remove the loop in JS that was changing the state using (state: index == items.rowSelected ? "active" : "default") but it didn’t work well, reason is when mouse is clicked index have the value of the current rectangle only, so no knowledge about past is there hence loop is needed.
If I keep it outside the mouse area then it works but only for first iteration at the time of initialisation but I need it to check whenever mouse is clicked.

By-mistake I closed the previous pull request, here is the link: https://github.com/bdoin/GCompris-qt/pull/125 

Now activity looks like this.
screenshot from 2016-03-28 01 23 09


bdoin:
When I resize the window's width, the instructions go on top of the table board
When the numpad is there the left one if over the table board
If I enter the result then click on the correct row/column the bonus does not pop up

On the code, try to stick to 80 column (especially in the checkit function).

There are errors in the console:
qrc:/gcompris/src/activities/multiplication-chart/MultiplicationChart.qml:219: TypeError: Cannot read property 'height' of null
qrc:/gcompris/src/activities/multiplication-chart/MultiplicationChart.qml:221: TypeError: Cannot read property 'width' of null
qrc:/gcompris/src/activities/multiplication-chart/MultiplicationChart.qml:60: TypeError: Cannot read property of null
qrc:/gcompris/src/activities/multiplication-chart/MultiplicationChart.qml:61: TypeError: Cannot read property of null

Beside this the code is clean but I question the value of this activity. You don't have to think that much to use it. We have to talk with Emmanuel (Allon) about that.

varun13169:
Hi, 
I'm not getting any errors in my console, may be they are because of main.height and main.width; will look into that.

For bonus to pop up, I specifically maintained the order according to instructions; I can incorporate the code for any arbitrary order as well. 

Allon told me that he used this Montessori activity to teach his students, although yes we can discuss with him to make it better :)

PR #129, opened 2016-03-29
Rahulyadav practice

rahulyadav170923:


bdoin:
Got a live review. Gave you several ideas to improve the UI. Need to see with Allon if adding boxes to represent numbers is a good idea.

PR #130, opened 2016-03-31
Port - PhotoHunter

StefanToncu:
Hi! I finished porting PhotoHunter from [gtk+](http://gcompris.net/screenshots-en.html#photohunter).
Please let me know what improvements i should add to the current version.

If someone wants to help with the animations, i will be very happy to accept.
[Phabricator task](https://phabricator.kde.org/T1535)


src/activities/photo_hunter/ActivityInfo.qml
```diff
@@ -0,0 +1,34 @@
...
+ */
+import GCompris 1.0
+
+ActivityInfo {
+  name: "photo_hunter/Photo_hunter.qml"
+  difficulty: 1
+  icon: "photo_hunter/photo_hunter.svg"
+  author: "Stefan Toncu &lt;stefan.toncu@cti.pub.ro&gt;"
+  demo: true
+  title: "Photo hunter"
```
petitlapin: use qsTr() for string that are translatable

src/activities/photo_hunter/ActivityInfo.qml
```diff
@@ -0,0 +1,34 @@
...
+  description: "Find the differences between the two pictures!"
+  //intro: "look at the two pictures carefully and click where you see a difference"
+  goal: "Visual Discrimination"
+  prerequisite: ""
+  manual: "Observe the two pictures carefully. There are some slight differences. When you find a difference you must click on it."
+  credit: ""
+  section: "fun"
+}
```
petitlapin: there is a createdInVersion: 6000 missing

src/activities/photo_hunter/Photo_hunter.qml
```diff
@@ -0,0 +1,176 @@
...
+            border.width: 2
+            border.color: "black"
+
+            color: "red"
+
+            GCText {
+                id: problemTxt
+                anchors.centerIn: parent
+                fontSize: items.cellSize / 7.5
+                text: qsTr("Click on the differences between the two images!   " + items.totalFound + "/" + items.total + " found.")
```
petitlapin: use qsTr("text %1").arg(variable) when you have variables. Also, avoid having lines bigger than 80 characters if possible

src/activities/photo_hunter/Photo_hunter.qml
```diff
@@ -0,0 +1,176 @@
...
+            height: parent.height - background.barHeight - 30
+
+            anchors.top: problem.bottom
+            anchors.horizontalCenter: parent.horizontalCenter
+            anchors.margins: 10
+
+            //left/top image
+            Observe {
+                id: img1
+                name: img2
```
petitlapin: Why having img1 name img2 and the opposite below?

src/activities/photo_hunter/resource/README~
```diff
@@ -0,0 +1,10 @@
+https://openclipart.org/detail/17626/castle
```
petitlapin: this temporary file should not be commited

src/activities/photo_hunter/Photo_hunter.qml
```diff
@@ -0,0 +1,176 @@
...
+
+            property int barHeightAddon: ApplicationSettings.isBarHidden ? 1 : 3
+            property int cellSize: Math.min(background.width / 11 , background.height / (9 + barHeightAddon))
+        }
+
+        onStart: { Activity.start(items) }
+        onStop: { Activity.stop() }
+
+        property bool vert: background.width < background.height
+        property double barHeight: (items.barHeightAddon == 1) ? 0 : 160
```
petitlapin: using bar.height instead of 160 does not work?

petitlapin:
Hi,

I have this error when finishing first level for first time: qrc:/gcompris/src/activities/photo_hunter/Photo_hunter.qml:99: TypeError: Cannot read property of null

Why do you have both levels in qml files and in photo_hunter.js? On my side, to add levels, it would be better to have them outside the main js file.

Johnny

StefanToncu:
**All the previous comments have been addressed**.
I also made _**2 changes**_: 
- added **particles effect** when clicking on a difference
- added a field in levels: "radius", which will scale the blue circle (that marks the differences) to the specified "radius". It is **visible in level 3**, when clicking on the zebra

petitlapin:
Hi,

when doing a bad click, maybe show a red cross?

Johnny

StefanToncu:
yep, the red cross is here now.
I also added some "good sounds" for pressing on the differences and "bad sound" when pressing on anything else (when showing the red cross).

src/activities/photo_hunter/Photo_hunter.qml
```diff
@@ -0,0 +1,172 @@
+/* GCompris - photo_hunter.qml
```
bdoin: Naming conventions implies Qml file to by CamelCase so this one must be PhotoHunter.qml

src/activities/photo_hunter/resource/images/photo9.svg
```diff
@@ -0,0 +1,1295 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
bdoin: You should put the images in resource/ not in a subdirectory

src/activities/photo_hunter/resource/README
```diff
@@ -0,0 +1,11 @@
+https://openclipart.org/detail/17626/castle
```
bdoin: We need to keep track of the reference on each resource file. For example we need to keep track of what the file photo1.svg refers to.

bdoin:
As the images takes a lot of disk space, we could be smarter and have a single image per level. It could be done by creating a single svg with tags to mark items with differences (svg is an xml format). Then at runtime we load the first image normally and change it to have the tagged items to appear. The second image is then loaded from the first image plus xml modification to make the changes visible.

src/activities/photo_hunter/Observe.qml
```diff
@@ -0,0 +1,145 @@
...
+                    id: particles
+                    clip: false
+                }
+            }
+
+            MouseArea {
+                id: mouseArea
+                anchors.fill: parent
+                onClicked: {
+                    img1.repeater.itemAt(index).particleLoader.active = active
```
bdoin: It would be better to call a function on the items instead of being intrusive here.

PR #131, opened 2016-04-02
Multiplication chart

varun13169:
Based on the feedback from previous PR, I made these improvements.

1) Added the support to select correct row, column or enter answer in any order; for bonus to show up.
2) Resized the activity so that numpad and multiplication board don’t overlap with each other.
3) Resizing issue resolved, now everything is dependent on background height and width.
4) No errors on console

here is the ref. to previous PR: https://github.com/bdoin/GCompris-qt/pull/127

Here is how the activity looks like:
screenshot from 2016-04-02 19 24 04


src/activities/multiplication-chart/MultiplicationChart.qml
```diff
@@ -0,0 +1,312 @@
...
+        onStop:  Activity.stop()
+
+
+
+        /////////////////////////////////////////////////////
+        Rectangle {
+            id: wrap
+            color: "transparent"
+            opacity: 1
+            height: background.height < background.width ? background.height : background.width
```
petitlapin: you can use: Math.min(background.height, background.width)

src/activities/multiplication-chart/MultiplicationChart.qml
```diff
@@ -0,0 +1,312 @@
...
+            width:  background.height < background.width ? background.height : background.width
```
petitlapin: you can set it equal to height

varun13169: Generally height is smaller than width, but this is to cover the case when window is resized and width becomes smaller than height. :)

petitlapin: Yes, as it is the same line as above you can put:
height: Math.min(background.height, background.width)
width:  height
you'll always have a square this way and no code repetition (it is also clearer for review)

varun13169: understood what wanted to say :) 
done

bdoin:
I tried it and I have many errors like this:
QFont::setPointSizeF: Point size <= 0 (0.000000), must be greater than 0

On my system with a higt definition screen all the texts are way too huge. You calculation does not take in account that font size are automatically adapted. In general, you should not play too much with font sizes just use one of the predefined one like largeSize if this is the one you want.

src/activities/multiplication-chart/MultiplicationChart.qml
```diff
@@ -0,0 +1,312 @@
...
+            id: instruction
+            anchors {
+                top: parent.top
+                topMargin: (wrap.height/4 - instruction.height/3)
+                left: grid.right
+                leftMargin: (wrap.width*17)/683 // 1366*17/683 = 34
+            }
+            fontSize: (wrap.height*7)/384  //768*7/384 = 14
+            text: qsTr("Instruction:\n")+
+                    qsTr("   Multiplicand x Multiplier\n")+
```
bdoin: This is not a nice way to align text. It may be aligned or not depending on the font;

src/activities/multiplication-chart/MultiplicationChart.qml
```diff
@@ -0,0 +1,312 @@
...
+        /////////////////////////////////////////////////////
+        Rectangle {
+            id: wrap
+            color: "transparent"
+            opacity: 1
+            height: Math.min(background.height , background.width)
+            width:  height
+            anchors.centerIn: background
+
+        Grid {
```
bdoin: Bad indentation.

src/activities/multiplication-chart/MultiplicationChart.qml
```diff
@@ -0,0 +1,312 @@
...
+                        text: (Math.floor(index/10) + 1) * (index%10 + 1)
+                        anchors.centerIn: parent
+                        fontSize: (wrap.height*6)/384  //768*6/384 = 12
+                    }
+
+                }
+            }
+        }
+
+        GCText{
```
bdoin: You need to set a width for the text item and have it being layout properly.

PR #135, opened 2016-04-10
Multiplication Chart

varun13169:
Made changes to make activity responsive


src/activities/ballcatch/Ballcatch.qml
```diff
@@ -34,6 +34,7 @@ ActivityBase {
 
     Keys.onPressed: {
         Activity.processKey(event)
+        console.log(ApplicationSettings.baseFontSize, "asdfg",  ApplicationInfo.fontRatio)
```
petitlapin: you should not have commited that

src/activities/multiplication-chart/ActivityInfo.qml
```diff
@@ -0,0 +1,43 @@
...
+  difficulty: 1
+  icon: "multiplication-chart/multiplication-chart.svg"
+  author: "Varun Kumar &lt;varun13169@iiitd.ac.in&gt;"
+  demo: true
+  title: qsTr("Multiplication Chart")
+  description: qsTr("Simple Multiplication of numbers from 1-10")
+  //intro: "put here in comment the text for the intro voice"
+  goal: qsTr("To multiply the given operands, select them on the board and state the answer")
+
+  prerequisite: qsTr("Simple Intro to multipliciation (not a necessity)")
```
petitlapin: multiplication

src/activities/multiplication-chart/MultiplicationChart.qml
```diff
@@ -0,0 +1,315 @@
...
+            property Item main: activity.main
+            property alias background: background
+            property alias bar: bar
+            property alias bonus: bonus
+            property alias repeaterGridRow: repeaterGridRow
+            property alias repeaterGridCol: repeaterGridCol
+            property alias repeater: repeater
+            /*variables*/
+            property int spacing: 5
+            property int rectHeight: wrap.height/25
```
petitlapin: if you don't use it in the js, you can remove it from the items object

src/activities/multiplication-chart/MultiplicationChart.qml
```diff
@@ -0,0 +1,315 @@
...
+            opacity: 1
+            height: Math.min(background.height , background.width)
+            width:  height
+            anchors.centerIn: background
+
+        Grid {
+            id:gridRow
+            anchors {
+                top: parent.top
+                topMargin: 20*(ApplicationInfo.ratio/1.476923076923077) //(wrap.height*5)/192     // 768*5/192 = 20
```
petitlapin: why 1.476923076923077?

varun13169: For my machine this ratio was 1.476923076923077
So, either I calculate (20/1.476923076923077)~13.3 and then multiply it it ApplicationInfo.ratio
or
do it this way 20*(ApplicationInfo.ratio/1.476923076923077)

I prefer the latter one just for my own ease, both will eventually result the same value.

bdoin: You should not code specifically for your screen. This 1.47 is non sense. Just do 15 * ApplicationInfo.ratio, we never need to be pixel perfect.

varun13169: done.

PR #136, opened 2016-04-13
Share

StefanToncu:
New activity - Share
Video preview: [Youtube/Share](https://youtu.be/HsWKnUhNOaA)
Phabricator task: [Share Activity](https://phabricator.kde.org/T1739)


src/activities/share/ActivityInfo.qml
```diff
@@ -0,0 +1,34 @@
...
+  demo: true
+  title: qsTr("Share the candies")
+  description: qsTr("Try to split the candies to a given number of children")
+  //intro: "put here in comment the text for the intro voice"
+  goal: qsTr("Learn the division of numbers")
+  prerequisite: qsTr("Know how to count")
+  manual: qsTr("Follow the instructions shown on the screen: first, drag the given number of boys/girls to the middle, then drag candies to each child's rectangle.")
+  credit: ""
+  section: "math"
+}
```
petitlapin: it misses a createdInVersion: 6000

src/activities/share/DropChild.qml
```diff
@@ -0,0 +1,188 @@
...
+
+Rectangle {
+    id: dropChild
+    width: items.cellSize * 3
+    height: items.cellSize * 4.5
+    color: "transparent"
+    z: 5
+
+    property string name
+    property alias childImg: childImg
```
petitlapin: use complete names for variable (childImage in this case).

src/activities/share/Share.qml
```diff
@@ -0,0 +1,399 @@
...
+                            (items.nCandies - Math.floor(items.nCandies / items.nChildren) * items.nChildren)) ?
+                               items.nCandies - items.background.nCrtCandies : 0
+
+
+        onRestChanged: {
+            //show message for rest if the rest is available and not already displayed on the board
+            if (rest > 0 && basketShown() === false) {
+                background.resetCandy()
+                instruction.opacity = 1
+                instruction.text = "now drag the basket in the center and place the remaining candies inside it"
```
petitlapin: missing qsTr()

src/activities/share/Widget.qml
```diff
@@ -0,0 +1,154 @@
+/* GCompris - Widget.qml
```
petitlapin: Do you have a better name than Widget which is quite generic?

src/activities/share/resource/README~
```diff
@@ -0,0 +1,5 @@
+Image source:
```
petitlapin: this file should not be commited

src/activities/share/share.js
```diff
@@ -0,0 +1,104 @@
...
+    items.basketWidget.canDrag = true
+    items.basketWidget.element.opacity = 0
+}
+
+function nextSubLevel() {
+    items.currentSubLevel ++
+    if (items.currentSubLevel === items.nbSubLevel) {
+        nextLevel()
+    }
+    setUp()
```
petitlapin: put an else here, or you'll do setUp once for nothing

StefanToncu:
The RefferenceError is now gone. I made the other necessary changes as well.
A new update: when dragging a candy from a rectangle back to the left/top widget panel, it is now displayed on top of the left/top widget (before the last commit, it went behind).

I'm waiting for more notes/observations on the activity.

src/activities/share/resource/board/board0.qml
```diff
@@ -0,0 +1,47 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+import QtQuick 2.0
+
+QtObject {
```
petitlapin: Is it possible to have dynamic levels? With some randomness, this way, children can play multiple times

StefanToncu: Of course it is possible. I will try to do it, but also keep in mind that the first levels should have a smaller difficulty.

petitlapin: Yes, for this you can probably play with bounds

petitlapin:
Hello,

I've just tested it. Not sure if it's best we talk here or by mail (I put my comments here)?

* If the number of children is restricted since the beginning of the level (there are **1** boy and **1** girl and we can only place **1** of each also), is there a reason that we don't directly place them? Or would it be better to allow to place as many as children they want?

* Same for the basket, if we know there is going to be a rest, can't we already place it (or at least put it visible since the beginning because it only appears when it's "good" now)?

* On ok click, I wouldn't reset everything, I would keep it like this.

* As said above, it would be great if the number of candies/childs were randomized (keeping difficulty levels using min/max bounds). This way, we can have more levels and a bigger life time duration.

Johnny

src/activities/share/WidgetOption.qml
```diff
@@ -0,0 +1,154 @@
...
+ *   This program is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+import QtQuick 2.5
```
bdoin: QtQuick 2.1 would be better.

bdoin:
* As Johnny mentioned, I agree that "on ok click (on a lost), I wouldn't reset everything, I would keep it like this."
* At level 3 sublevel 2 and probably others, If I don't pick the basket when suggested I cannot drag it anymore (I put another candy instead of the basket when suggested).
* I am fine with the increasing difficulty and I don't suggest to put random here or we risk missing a normal progression
* I am fine with have to put the boy/girl/basket if Allon suggest it is part of the pedagogical strategy.
* No to be done now but we could make more levels. Like:

* When children are placed they already contains some candies so the children cannot just distribute candies one by one, he must count.
* Remove the number on the children drop area to force counting.
* Place candies on random spot in the children area to make it harder to just see the solution without counting.
* We could say 'give as many candies to bob than alice'. And when Alice is placed she has already some candies

src/activities/share/share.js
```diff
@@ -0,0 +1,240 @@
...
+    items.dataset.source = filename
+
+    setUp()
+}
+
+function setUp() {
+    var levelData = items.dataset.item
+
+    // use board levels
+    if (currentLevel < 7) {
```
petitlapin: I would use a boolean property in the dataset instead of hardcoding this value.

src/activities/share/share.js
```diff
@@ -0,0 +1,240 @@
...
+
+        if (items.totalBoys > 1)
+            boys = "boys"
+        else boys = "boy"
+
+        if (items.totalGirls > 1)
+            girls = "girls"
+        else girls = "girl"
+
+        items.instruction.text = qsTr("Place " + items.totalBoys + " " + boys + " and " +
```
petitlapin: you should not handle translation like that here.
It should be like in https://github.com/gcompris/GCompris-qt/blob/master/src/activities/balancebox/editor/balanceboxeditor.js#L157

src/activities/share/Share.qml
```diff
@@ -0,0 +1,461 @@
...
+
+        GCText {
+            id: wrongMoveText
+            anchors.horizontalCenter: grid.horizontalCenter
+            anchors.verticalCenter: parent.verticalCenter
+            fontSize: background.vert ? hugeSize : largeSize
+            opacity: wrongMove.opacity
+            z: wrongMove.z
+            color: "#404040"
+            text: qsTr(" Too many candies!! ")
```
petitlapin: You can remove the spaces at beginning and end and one exclamation mark.

PR #137, opened 2016-04-13
MirrorGame version 2.0

shubham-1996:
Whole new version of mirrorgame activity with dynamic light path checking system.Although,still a few corner cases are left to deal with but the algorithm works perfectly.The graphics and animation have to be added bu the over all skeleton of the game is ready.Please provide feedbacks.


src/activities/mirrorgame/ActivityInfo.qml
```diff
@@ -0,0 +1,35 @@
+/* GCompris - ActivityInfo.qml
+ *
+ * Copyright (C) 2016 Shubham Nagaria shubhamrnagaria@gmail.com
```
petitlapin: Use < > to surround the email.

src/activities/mirrorgame/ActivityInfo.qml
```diff
@@ -0,0 +1,35 @@
...
+ */
+import GCompris 1.0
+
+ActivityInfo {
+  name: "mirrorgame/Mirrorgame.qml"
+  difficulty: 3
+  icon: "mirrorgame/mirrorgame.svg"
+  author: "Shubham Nagaria &lt;shubhamrnagaria@gmail.com&gt;"
+  demo: true
+  title: "Mirror Game"
```
petitlapin: use qsTr() for translatable strings

src/activities/mirrorgame/BulbOff.qml
```diff
@@ -0,0 +1,33 @@
+/* GCompris - BulbOn.qml
```
petitlapin: it should be BulbOff.qml

src/activities/mirrorgame/README.md
```diff
@@ -0,0 +1 @@
+
```
petitlapin: If the file is empty, you don't need it

src/activities/mirrorgame/mirrorgame.js
```diff
@@ -0,0 +1,536 @@
...
+            }
+
+        }
+    checkLightPath();
+    }
+
+function mod(a,b)
+{   return (a>=b)? (a-b):(b-a);}
+
+function min(a,b)
```
petitlapin: There is Math.min and Math.max

src/activities/mirrorgame/mirrorgame.js
```diff
@@ -0,0 +1,536 @@
...
+
+function createLightgrid()
+{
+    for(var x = 0;x < items.rows; x++) {
+         for(var y = 0;  y < items.columns; y++){
+             for(var w = x+1; w < items.rows; w++){
+                 for(var z = y+1; z < items.columns; z++){
+                     if((x-w) === (y-z) && x!==w && y!==z)
+                     {
+                 var component = Qt.createComponent("Light-45.qml");
```
petitlapin: you can probably just create once the component at beginning.

src/activities/mirrorgame/mirrorgame.js
```diff
@@ -0,0 +1,536 @@
...
+                                                       "visible" : false,
+                                                        "uniquename" : "lightpath" +x+y+w+z
+                                                      }); }}
+
+}}}}
+
+    for( x = 0;x < items.rows; x++) {
+         for( y = 1;  y < items.columns; y++){
+             for( w = x+1; w < items.rows; w++){
+                 for( z = y-1; z >=0; z--){
```
petitlapin: the code looks a lot like above, you can probably factor it inside of doing twice all the loops

src/activities/mirrorgame/mirrorgame.js
```diff
@@ -0,0 +1,536 @@
...
+{
+    for(var x = 0;x < items.rows; x++) {
+         for(var y = 0;  y < items.columns; y++){
+             for(var w = x+1; w < items.rows; w++){
+                 for(var z = y+1; z < items.columns; z++){
+                     if((x-w) === (y-z) && x!==w && y!==z)
+                     {
+                 var component = Qt.createComponent("Light-45.qml");
+                 if (component.status === 1) {
+                 Light[count++] = component.createObject(items.grid,{
```
petitlapin: Instead of having a file for all, I thik you can create an Image directly and set the source depending on the Item you want to create

src/activities/mirrorgame/mirrorgame.js
```diff
@@ -0,0 +1,536 @@
...
+
+    } else {
+        console.log("error loading block component");
+        console.log(component.errorString());
+        return false;
+    }
+    return true;
+}
+function createTorch(column, row) {
+
```
petitlapin: same code as above, you can simplify it

src/activities/mirrorgame/mirrorgame.js
```diff
@@ -0,0 +1,536 @@
...
+                                                        "width":items.cellSize,
+                                                        "height":items.cellSize,
+                                                        "uniquename" : "bulboff" + row+column
+                                                       });
+        }
+
+    return true;
+}
+function checkLightPath()
+{var i,j,k
```
petitlapin: fix indentation

src/activities/mirrorgame/mirrorgame.js
```diff
@@ -0,0 +1,536 @@
...
+
+
+for(var z = 0;z<numberofmirrors;z++){
+  for(i=0;i<5;i++) {
```
petitlapin: why all these 5? If it is the same meaning, create a variable

src/activities/mirrorgame/resource/README
```diff
@@ -0,0 +1,3 @@
+Light source and mirrors are made using Inkscape.
+Bulbs are taken from openclipart.org.
```
petitlapin: you need to give the links of the images

src/activities/mirrorgame/resource/README.cd
```diff
@@ -0,0 +1 @@
+
```
petitlapin: empty file?

src/activities/mirrorgame/resource/bar_ok.svg
```diff
@@ -0,0 +1,83 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
petitlapin: you don't need to copy the one from core, you can use it directly

src/activities/mirrorgame/mirrorgame.js
shubham-1996: i have created the light grid once only...for the grid part i have to add mirrors simultaneously with grid ,that's why i am creating it at every level

src/activities/mirrorgame/mirrorgame.js
shubham-1996: for dynamic creation of objects it needs a qml file

petitlapin:
Hi,

I would see this activity more in the fun than in experiment category.

A few comments:
* Once you have found the good solution but didn't press ok, if you change a mirror (thus, the solution is no more good) but click on OK, then the solution is marked as good where it should not be.

* When resizing the screen, the mirrors, light are not replaced well.

* If you click twice in a row (before the end of the rotation) in a mirror, it does not place well (but the light is the one awaited).

* If you go on the activity, leave it and go in it again, there is an error (qrc:/gcompris/src/activities/mirrorgame/mirrorgame.js:129: TypeError: Cannot read property 'uniquename' of undefined) and the game is not playable at all.

* Also, you don't need to do one commit per file, it's more one commit per functionality (with commit message something like: "activity, change done" (replace activity,with the name of the updated activity).

JohnnyJ

src/activities/mirrorgame/Mirrorgame.qml
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+
+        Grid {
+            id: grid            
+            anchors {
+            verticalCenter: parent.verticalCenter
+            horizontalCenter: parent.horizontalCenter
+            }
+            columns: 5
```
petitlapin: Is it possible to make the grid columns and rows numbers not fixed? This way, we can have easy levels with only a 3*3 square (quite easy) and harder levels with more than a 5*5 square.

src/activities/mirrorgame/Mirrorgame.qml
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        Image{
+            id:ok
+            source:"qrc:/gcompris/src/core/resource/bar_ok.svg"
+            anchors.horizontalCenter: background.horizontalCenter
+            visible: false
+            MouseArea{
+                anchors.fill: parent
+                onClicked: {Activity.win()
```
petitlapin: be careful of indentation.
Leave a space between the property and its value:
source: "...".
Same for an object declaration:
MouseArea {
Return to the line after a { if it takes more than one line (if it is readable, you can do { "..." }).

src/activities/mirrorgame/mirrorgame.js
```diff
@@ -0,0 +1,526 @@
...
+            [//level5
+             [0,0,0,0,0],
+             [1,0,0,0,3],
+             [0,2,0,0,0],
+             [2,0,0,0,0],
+             [0,2,0,0,0]
+            ]
+                    ]
+var currentLevel = 0
+var numberOfLevel = 5
```
petitlapin: you don't need this, the number of levels is data.length

src/activities/mirrorgame/mirrorgame.js
```diff
@@ -0,0 +1,526 @@
...
+var arr
+var items
+var count = 0
+var num = 0
+var numberofmirrors = 0
+var i,j
+var Light
```
petitlapin: use lower case as first letter

src/activities/mirrorgame/mirrorgame.js
```diff
@@ -0,0 +1,526 @@
...
+             [2,0,0,0,0],
+             [0,2,0,0,0]
+            ]
+                    ]
+var currentLevel = 0
+var numberOfLevel = 5
+var arr
+var items
+var count = 0
+var num = 0
```
petitlapin: it is not explicit on what it does

src/activities/mirrorgame/mirrorgame.js
```diff
@@ -0,0 +1,526 @@
...
+var numberofmirrors = 0
```
petitlapin: use camelCase for variable names (numberOfMirrors)

src/activities/mirrorgame/mirrorgame.js
```diff
@@ -0,0 +1,526 @@
...
+            if(arr[y][x]===3)
+            {createBulb(x,y);}
+
+            }
+
+        }
+    checkLightPath();
+    }
+
+function mod(a,b)
```
petitlapin: is it the same asMath.abs(a-b)?

src/activities/mirrorgame/mirrorgame.js
```diff
@@ -0,0 +1,526 @@
...
+         bulbon[num] = component5.createObject(items.grid,{"x":column * items.cellSize+items.grid.spacing,
+                                                    "y":row * items.cellSize+items.grid.spacing,
+                                                    "width":items.cellSize,
+                                                    "height":items.cellSize,
+                                                    "uniquename" : "bulbon" + row+column,
+                                                    "visible":false
+                                                   });
+
+    }
+        if (component6.status === 1) {
```
petitlapin: you can probably have only one bulb array and lighten it when necessary (changing the image source)

src/activities/mirrorgame/mirrorgame.js
```diff
@@ -0,0 +1,526 @@
...
+                          done = 1
+                      }
+                  }}}
+
+
+              }}if(done)break}if(done)break}
+              break
+
+          case 2:
+              while(currLight[i][j].length!==0)
```
petitlapin: what are the differences between the different cases in the switch.
Maybe some code is factorisable?

src/activities/mirrorgame/mirrorgame.js
```diff
@@ -0,0 +1,526 @@
...
+                                && a<i && b<j)
+                           {
+                            Light[c].visible = true;
+                            currLight[i][j].push(Light[c])
+                        }
+                }}}
+                }
+                }if(done)break}if(done)break}
+                break
+    }
```
petitlapin: indentation. We don't know which { corresponds to which }

src/activities/mirrorgame/mirrorgame.js
```diff
@@ -0,0 +1,526 @@
...
+                   for(var e = 0;e < count;e++){
+                    for(var f = 0;f < num;f++){
+                        if(bulboff[f].uniquename === "bulboff"+a+b &&
+                                (Light[e].uniquename === "lightpath"+a+b+c+d||
+                                 Light[e].uniquename === "lightpath"+c+d+a+b)
+                               &&
+                                Light[e].visible === true)
+                        {
+                            bulboff[f].visible = false;
+                            bulbon[f].visible = true;
```
petitlapin: this goes in pair with the comment: "you can probably have only one bulb array and lighten it when necessary (changing the image source)"

petitlapin:
hi, I'm closing it for now as it's been a long time since the last change.
Johnny

PR #139, opened 2016-04-16
Baby processor

divyam3897:
Changed the background color to green and text color to white.Its now like a blackboard with white chalk with all buttons at the top right visible.


petitlapin:
Hi,

you should have done a separate branch for this fix created from master (so without your other changes on another activity).

Anyways, yesterday, after discussion with Animtim, we decided to revert the change on the buttons except for chess, he'll apply a patch next week.

JohnnyJ

divyam3897:
Hey,
I did create a seperate branch but by mistake I created that branch to my
previous branch instead of master which caused this error.
Anyways If we  don't revert the button changes doesn't this change seems
good? The buttons are completely visible now and seem pretty good and
working.

Thank you
Regards
Divyam Madaan

PR #141, opened 2016-04-28
Up to date Photo-Hunter

stefant29:
the easiest way to do a clean PR was to create a new account


bdoin:
No the easiest way is not to create a new account. It is true that git can be hard to master but it it worth trying. For instance your github account becomes more important than a CV in a developer's career. You should concentrate on a single one. By the way, your contribution is very nice and has just been integrated in GCompris master.

PR #143, opened 2016-06-28
Created demo activity DRAWLETTERS(WIP) for review

Nitish18:
Activity Name :- nitish_drawletters
section :- fun
Description :- Children have to connect dots correctly to draw some letters & numbers.

Note:- The Activity only contains some prototype levels for review purpose.(WIP)


bdoin:
Just tested it. This is a good start and could be interesting as is to enhance the current drawnumber activity. But this is not what I have in mind for the letter drawing. It should let the children draw the letter, not the 'bold shape' of it. For instance to draw an O I expect a circle, not 2 circle to represent the inner an outer pert of the O.

When you pick images take care of the license and include the reference in the README. In general best is to stick to openclipart.org.

Nitish18:
Ok sir , I will keep in mind the changes you have mentioned above.
One more thing I want to ask , for example according to you to draw an O you expect a circle, not 2 circle to represent the inner an outer pert of the O , **so to draw it whether assistance dots would be provided to the children or not ??**
Regarding images, I had taken care of license before using it.

bdoin:
The idea is to provide many dots along the O path with no gap between them. The children will see the O drawn in dots that touch each others. For each image put the link and license in the README file.

PR #145, opened 2016-07-05
Images errors in lang resolved

divyam3897:
Resolved the errors libpng warning: iCCP: known incorrect sRGB profile in lang


petitlapin:
From Divyam:
http://stackoverflow.com/questions/22745076/libpng-warning-iccp-known-incorrect-srgb-profile
mogrify *.png 
if it works fine, we also need to process words.rcc

petitlapin:
HI,

it is pushed in kde repo.

Thanks,

Johnny

PR #146, opened 2016-07-06
Modified DrawLetters Activity

Nitish18:
Modified the "DrawLetters" Activity with following changes :-
1.) Changed name to drawletters instead of nitish_drawletters. 
2.) Added sound of letters at start & at end.
3.) Added sound when the dots are clicked.
4.) Added some color to the next dot to be clicked.
5.) Changed the images of each level.
6.) Added some more levels.
7.) Added instruction at bottom of each level.

Waiting for reviews & suggestions.


src/activities/drawletters/ActivityInfo.qml
```diff
@@ -0,0 +1,35 @@
...
+ */
+import GCompris 1.0
+
+ActivityInfo {
+  name: "drawletters/Drawletters.qml"
+  difficulty: 1
+  icon: "drawletters/drawletters.svg"
+  author: "Nitish Chauhan &lt;nitish.nc18@gmail.com&gt;"
+  demo: true
+  title: "Draw Letters"
```
petitlapin: you need to add qsTr() there so the title will also be translated.

src/activities/drawletters/Drawletters.qml
```diff
@@ -0,0 +1,287 @@
...
+            property alias pointImageRepeater: pointImageRepeater
+            property alias segmentsRepeater: segmentsRepeater
+            property alias imageBack: imageBack
+            property int pointIndexToClick
+
+
+                    }
+
+
+        onStart: {
```
petitlapin: Globally, please remove all useless line breaks and indent the code correctly (brackets...).

src/activities/drawletters/Drawletters.qml
```diff
@@ -0,0 +1,287 @@
...
+            anchors.top: parent.top
+            width: background.width
+            height: background.height
+
+        }
+
+        GCText {
+                id: number
+                opacity: pointImageOpacity
+                text: "Connect the Dots in the right sequence to Draw"
```
petitlapin: qsTr() missing

src/activities/drawletters/Drawletters.qml
```diff
@@ -0,0 +1,287 @@
...
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+import QtQuick 2.2
+import GCompris 1.0
+import "../../core"
+import QtQuick.Particles 2.0
```
petitlapin: do you really need to import Particles?

src/activities/drawletters/Drawletters.qml
```diff
@@ -0,0 +1,287 @@
...
+                        }
+                    }
+                }
+
+            }
+
+            onPressed: {
+                checkPoints(touchPoints)
+                // to play upon clicking
+                activity.audioEffects.play(Activity.url + "buttonclick."+ApplicationInfo.CompressedAudio)
```
petitlapin: you can use "buttonclick.$CA" instead of "buttonclick."+ApplicationInfo.CompressedAudio

src/activities/drawletters/drawletters.qrc~
```diff
@@ -0,0 +1,48 @@
+<RCC>
```
petitlapin: don't commit this file

Nitish18: Hi, thanks for your suggestions . I want to ask one thing , I do not have to commit which file 1.) drawletters.qrc  or README or both ??

petitlapin: You commited src/activities/drawletters/drawletters.qrc~ and src/activities/drawletters/resource/README~ which are temporary files (the ~ at the end). Both should be deleted (but not README).
drawletters.qrc should also not be commited (it is not) because it is generated, don't try to commit it ;)

Nitish18: I have committed the changes you have told. Can you please tell me , if I have made a mistake again or not ??. Thanks.

src/activities/drawletters/ActivityInfo.qml
```diff
@@ -23,13 +23,10 @@ ActivityInfo {
   icon: "drawletters/drawletters.svg"
   author: "Nitish Chauhan &lt;nitish.nc18@gmail.com&gt;"
   demo: true
-  title: "Draw Letters"
+  title: qsTr("Draw Letters")
   description: qsTr("You have to join the dots in the right order to finally draw some letters & numbers")
-  //intro: "put here in comment the text for the intro voice"
   goal: qsTr("Children will learn about different letters & Numbers in a fun way.")
-  prerequisite: ""
```
petitlapin: I think these are mandatory (even empty) else it won't start (I can't test right now, so I'm just guessing)

src/activities/drawletters/drawletters.qrc~
petitlapin: Hi,
I can't test for now, I'll test as soon as I can. Globally, the code seems good.
Johnny

petitlapin:
Hi,

as discussed in irc, instead of having a new activity, you can inherit from drawnumber activity (like the clickanddraw activity). Changing the bool 'clickanddrawflag' with a string that will have 3 possible values depending on which activity you are: "clickanddraw", "drawnumber", "drawletter". Then, handle it in the js file to play the sound if you are in your mode (don't break the existing activities :)).

For the dots, I would put a higher z to the next items to select (I mean, if you have 1, 2, 3 that overrides, we should have the 1 in top, then the 2 then the 3).

I would put this activity in reading section (in ActivityInfo.qml).

if you resize the screen, the text is behind the bar and truncated (I'm not even sure the text is needed).

For the dots, I would put a higher z to the next items to select (for the A for example, the 19 hides the 2 but 2 would be the next one to select).

Thanks for your PR.

Johnny

PR #152, opened 2016-07-18
Drawletters

Nitish18:
Made the changes as told :-
1.) .wav file instead of .ogg sound file.
2.) some minor changes.(removed temporary files :) )
3.) to reduce the size of the images , I will implement this modification in my next Activity "Drawing numbers" . I will use two images , one for background & the other only for number.

Thanks for your feedback Johnny.
P.S :- If there is anything left , I will be happy to modify it.


petitlapin:
Hi Nitish, for future commits, there will be an issue with your author name: KDE needs to have two parts, "first name" "last name" (git commit --amend --author="Author Name <email@address.com>")
For me, this PR is good to be integrated, I'm waiting a few days to see if someone else has something to say, else I'll integrate it.

Thank you,

Johnny

Nitish18:
Thanks Johnny , next time I'll keep these things in mind.

petitlapin:
Also, as mentioned by dmadaan in the chan, it could be interesting to have the remaining letters (not necessarely in this PR but after the size reduction?) and lowercase letters?
It may be difficult for some letters if we have to "lift the pencil" (for f for example, we have to go backwards after the top to draw the middle line)?

petitlapin:
Hi,

I've merged it in master.
Thanks,

Johnny

PR #156, opened 2016-10-01
Background music

divyam3897:
Background Music added but still following issues left to be solved:
1. Getting the following error when download is completed: 
Download error 299: Checksum of downloaded file does not match:  "/home/divyam/.cache/KDE/gcompris-qt/data2/backgroundMusic-ogg.rcc". Though the download is completed successfully but as a result .rcc is not registered properly and the checkbox doesn't get checked (even after music is downloaded). Thus it downloads the music even when its already present.
2. The music doesn't play just after the download (after close of config). Though I added backgroundMusic in stack where voice and effects were loaded in main.qml.


src/activities/menu/ConfigurationItem.qml
```diff
@@ -404,6 +413,58 @@ Item {
                 }
             }
         }
+        
+        Flow {
```
hkaelber: Not sure if we need this row in the configuration. We don't  have it for the wordset. Maybe we handle it silently and under the hood. Not sure about it, let's keep it for now unless Johnny has an opinion.

src/core/ActivityInfoTree.cpp
```diff
@@ -290,6 +290,10 @@ void ActivityInfoTree::init()
                                    "/data2/" + QString("full-%1.rcc").arg(COMPRESSED_AUDIO)))
         qDebug() << "Registered the pre-download " << QString("full-%1.rcc").arg(COMPRESSED_AUDIO);
 
+    if(QResource::registerResource(QStandardPaths::writableLocation(QStandardPaths::CacheLocation) +
+                                   "/data2/" + QString("backgroundMusic-%1.rcc").arg(COMPRESSED_AUDIO)))
+        qDebug() << "Registered the pre-download " << QString("backgroundMusic-%1.rcc").arg(COMPRESSED_AUDIO);
+
     qmlRegisterSingletonType<QObject>("GCompris", 1, 0, "ActivityInfoTree", menuTreeProvider);
```
hkaelber: This does not belong here. Background-music is not pre-downloaded, it is downloaded dynamically as is wordset. Drop it here and start download in checkWordset() in main.qml via DownloadManager

src/core/ApplicationInfo.cpp
```diff
@@ -94,7 +94,9 @@ ApplicationInfo::ApplicationInfo(QObject *parent): QObject(parent)
 #endif
     // Get fonts from rcc
     const QStringList fontFilters = {"*.otf", "*.ttf"};
+    const QStringList backgroundMusicFilters = {QString("*.%1").arg(COMPRESSED_AUDIO) };
     m_fontsFromRcc = QDir(":/gcompris/src/core/resource/fonts").entryList(fontFilters);
+    m_backgroundMusicFromRcc = QDir(":/gcompris/data/backgroundMusic").entryList(backgroundMusicFilters);
```
hkaelber: This is too static. Either m_backgroundMusicFromRcc has to be updated whenever the resource is registered. Maybe even better: don't hold it in a member-variable just make ApplicationInfo::getBackgroundMusicFromRcc() return dynamically QDir(":/gcompris/data/backgroundMusic").entryList(backgroundMusicFilters); if it works

src/core/ApplicationInfo.cpp
```diff
@@ -183,6 +185,11 @@ QStringList ApplicationInfo::getFontsFromRcc()
     return m_fontsFromRcc;
 }
 
+QStringList ApplicationInfo::getBackgroundMusicFromRcc()
+{
+    return m_backgroundMusicFromRcc;
+}
```
hkaelber: cf. above:
return QDir(":/gcompris/data/backgroundMusic").entryList(backgroundMusicFilters);

src/core/ApplicationSettings.cpp
```diff
@@ -311,7 +320,6 @@ void ApplicationSettings::setIsAutomaticDownloadsEnabled(const bool newIsAutomat
         emit automaticDownloadsEnabledChanged();
     }
 }
-
```
hkaelber: Avoid such unrelated changes!

src/core/DownloadManager.cpp
```diff
@@ -153,6 +153,11 @@ QString DownloadManager::getVoicesResourceForLocale(const QString& locale) const
             .arg(ApplicationInfo::getInstance()->getVoicesLocale(locale));
 }
 
+QString DownloadManager::getBackgroundMusicResources() const
```
hkaelber: Not sure if we need such a function. Anyway lets keep it.

src/core/main.qml
```diff
@@ -47,121 +47,152 @@ Window {
...
+            backgroundMusic.stop()
+        }
+    }
 
     GCAudio {
         id: audioVoices
         muted: !ApplicationSettings.isAudioVoicesEnabled
-
+        
         Timer {
```
hkaelber: You added a lot of crap with such changes, i.e. inserted blanks at the beginning of the lines. Pls. don't, pls. revert these lines!
Hint: use qtcreator, which (configurably) removes trailing spaces/tabs upon save!

src/core/main.qml
```diff
@@ -47,121 +47,152 @@ Window {
...
         audioVoices.append(ApplicationInfo.getAudioFilePath("voices-$CA/$LOCALE/intro/" + name + ".$CA"))
     }
-
+    
     function checkWordset() {
-        var wordset = ApplicationSettings.wordset
-        if(wordset == '')
-            // Maybe the wordset has been bundled or copied manually
-            // we have to register it if we find it.
-            wordset = 'data2/words/words.rcc'
```
hkaelber: Why did you drop that?
Probably should do the same for backgroundMusic*.rcc.

src/core/main.qml
```diff
@@ -47,121 +47,152 @@ Window {
...
-                    if(wordset === arguments[0]) {
-                        DownloadManager.resourceRegistered.disconnect(arguments.callee);
-                        // force configuration to use the local wordset
-                        ApplicationSettings.wordset = wordset
-                    }
-                })
-            }
-
-
-        } else if(ApplicationSettings.wordset) { // Only if wordset specified
```
hkaelber: Same question: why did you change that?
As last said on IRC: I'd propose to handle backgroundMusic equivalently to wordset here (and rename checkWordset() to checkDataResources().
In the longer run I'd propose to factor all this code out to a ResourceManager. But this is out of scope for your task.

src/core/main.qml
```diff
@@ -47,121 +47,152 @@ Window {
...
                         qsTr("No"), null,
                         function() { pageView.currentItem.focus = true }
-            );
+                        );
         }
     }
     ChangeLog {
-       id: changelog
+        id: changelog
     }
```
hkaelber: Another unrelated blank.

src/core/main.qml
```diff
@@ -188,14 +219,18 @@ Window {
...
         else {
             // Register voices-resources for current locale, updates/downloads only if
             // not prohibited by the settings
             if (!DownloadManager.areVoicesRegistered()) {
                 if (DownloadManager.updateResource(
                             DownloadManager.getVoicesResourceForLocale(ApplicationSettings.locale))) {
+                    DownloadManager.downloadFinished.connect(function() {
+                        checkWordset();
+                        DownloadManager.downloadFinished.disconnect(arguments.callee);
+                    });
```
hkaelber: This connect() it not necessary, because we trigger checkWordset() on DownloadManager.resourceRegistered() one line below. Only one of the two is necessary

src/activities/menu/ConfigurationItem.qml
petitlapin: If I'm not wrong for wordset, the behavior is to check the box in config and then we use the complete wordset (note aside, it seems buggy for now as the only moment where we can actually download the rcc is at start up and if we check the box and don't have dataset we will have error and going to the activities, so for me we should ask to download the rcc if not present and if the user does not want, we keep the embedded wordset).
What you propose would be to download them (or ask user if he wants to download) when we check the "Enable background music" if they are not present (and we can probably go further as not showing it if we are not connected to network)?

petitlapin:
It seems I can't directly answer to your comment:

> Hi,
>two major issues (besides the one commented inline):
>1. It's wrong to register the backgroundMusic in ActivityInfoTree::init(). It has to be checked during >startup and probably if the enableBackgroundMusic config setting is activated (for the first time).
>2. Your Contents file on the server is wrong. It contains the wrong content. After having run >generate_backgroundMusic_rcc.sh you have to copy everything below .rcc/ to your upstream server, >including the generated Contents file, which contains after I ran it here:
>4ab15d3571700483d11a984f91315246 backgroundMusic-ogg.rcc

I generated it and uploaded it, I probably did something wrong regarding your comment ;), I'll regenerate it and upload it.

>This is the MD5 checksum of the generated rcc-file, which is checked by downloadManager against the >actual md5 checksum of the downloaded file to be able to detect incomplete downloads.
>Btw.: Did you also package other music-rccs for the other audio codecs. IIRC we need mp3 for apple >devices and aac for windows. (Correct me if I'm wrong, Johnny).

No you are totally right, I just did the ogg rcc to test, I'll need to create and upload the other once we have merged it.

>
>But you're close. Fix these issues and resend the pull-request. For the next round.

Holger (gandhi_m)

src/activities/menu/ConfigurationItem.qml
hkaelber: You're right, maybe we should change the behavior for wordset the same way:

If activated in config
    if rcc already downloaded (DownloadManager.haveLocalResource())
        update and/or register (DownloadManager.updateResource())
    else
        Popup dialog and ask for immediate download; if confirmed
            start download unconditionally (DownloadManager.downloadResource())
   
For backgroundMusic this would be triggered on enabledBackgroundMusic. For wordset (which is not your job for your pull request) on useExernalLargeWordset.

Alternatively we could skip the request dialog and simply download based on the automaticDownload setting.

src/core/main.qml
```diff
@@ -47,121 +47,152 @@ Window {
...
+    }
 
     function playIntroVoice(name) {
         name = name.split("/")[0]
         audioVoices.append(ApplicationInfo.getAudioFilePath("voices-$CA/$LOCALE/intro/" + name + ".$CA"))
     }
-
+    
     function checkWordset() {
-        var wordset = ApplicationSettings.wordset
```
hkaelber: Actually I suggested on IRC to do the initial updating/downloading/registering of backgroundMusic in this function equivalently to wordset.

src/core/main.qml
```diff
@@ -47,121 +47,152 @@ Window {
...
                 audioVoices.append(ApplicationInfo.getAudioFilePath("voices-$CA/$LOCALE/misc/welcome.$CA"));
             }
         }
-
+        
         Component.onCompleted: {
-            if(ApplicationSettings.isAudioEffectsEnabled)
-                append(ApplicationInfo.getAudioFilePath("qrc:/gcompris/src/core/resource/intro.$CA"))
+            if(!ApplicationSettings.isAudioEffectsEnabled && isBackgroundMusicAllowed) {
+                 Math.floor(Math.random()*rccBackgroundMusic.length);
```
divyam3897: No, the background music was played only when audio effects was enabled earlier but there can be a case when the user doesn't want to keep the audio effects but wants to play the music? So, I just kept it as a check if audio effects is disabled then also the user can play the music.

src/activities/menu/ConfigurationItem.qml
divyam3897: So, what could be the optimum choice keeping in mind the advantages and disadvantages of both the methods? Should we change it the wordset way and do it as done in checkWordset() or keep it the way it is currently (letting the user decide when to download and update when necessary and not just keeping the download for start)?

src/core/main.qml
divyam3897: Apologies for it! I will surely modify all the changes in the next commit. I don't know if Qt does that when I tried indenting once using its ctrl + I it pretty much messed up the indentation and I had to revert all the changes back

src/core/main.qml
divyam3897: No changes were required in this function. I will revert them back. I was not sure if that is required to handle it equivalently to wordset here ? Isn't the work done by the Download Manager in the check update that we created? I thought that it would do the job for checking if we already have the .rcc and if not then it would start its job for downloading?

src/core/main.qml
divyam3897: I didn't do these changes ;) It was there already for voices and I think that it is required there?

src/activities/menu/ConfigurationItem.qml
petitlapin: I agree with Holger. If automatic download, we don't show the dialog and start the download else we show it. At the end of the download, we enable/register the music.
Do not touch the wordset for now, it should be done in another pr

src/activities/menu/ConfigurationItem.qml
```diff
@@ -404,6 +413,56 @@ Item {
...
+            
+            Button {
+                id: backgroundMusicButton
+                height: 30 * ApplicationInfo.ratio
+                visible: ApplicationInfo.isDownloadAllowed
+                text: backgroundMusicRow ? qsTr("Check for updates") :
+                qsTr("Download")
+                style: GCButtonStyle {}
+                onClicked: {
+                if(DownloadManager.downloadResource(DownloadManager.getBackgroundMusicResources())) {
```
petitlapin: there, you download without checking if it was already present. If it is already present, you should just enable the sounds. Anyway, this button will disappear, it will be the code for the enableBackgroundMusicBox GCCheckBox.

src/core/main.qml
petitlapin: what I meant by useless is that you don't assign the result of the floor().

src/core/main.qml
petitlapin: If they are in the PR, then you probably did some change?
I tested it and if we don't have the wordset, then the pop up to download them at start up is displayed twice while with the master branch code, it is displayed only once

src/tools/generate_backgroundMusic_rcc.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
```
petitlapin: you should not have this file, it is already in tools/

src/core/main.qml
hkaelber: Correct, DownloadManager's task is to download, update and register rcc-files. But you have to tell him when and how. Use updateResource() when you want him to check for updates,  download and register them if yes and simply register the existing resource if there are no updates available. Use downloadResource() if you're sure you want to download and are allowd to (e.g. after you requested a user's confirmation, because this method does not check the "automaticDownloadAllowed" setting.
But I agree, that checkWordset() is really ugly. Let's keep it for now like that and add downloading backgroundMusic equivalentlu, and I put on my roadmap to facilitate handling registered data resources.

src/activities/menu/ConfigurationItem.qml
divyam3897: There is no automatic download for now which is basically start to download when we start GCompris right? Currently the button only does the job. Can it be done that if the music is enabled and there is no rcc then we download the rcc and get it registered else if the rcc is not registered then get it registered? Do you see it this way only?

PR #157, opened 2016-10-04
redraw activity: improve color list bar

aseemarora1995:


src/activities/redraw/Redraw.qml
```diff
@@ -99,120 +100,127 @@ ActivityBase {
...
+                height: background.height
+                boundsBehavior: Flickable.StopAtBounds
+                contentHeight: items.numberOfColor * width
+                bottomMargin: bar.height
+                Column {
+                    id: colorSelector
+                    Repeater {
+                        model: items.numberOfColor
+                        Item {
+                            width: 70 * ApplicationInfo.ratio
```
petitlapin: I would put an id to the Flickable and use flickable.width there. This way, if we change the flickable width for some reason, it will be the same here

src/activities/redraw/Redraw.qml
```diff
@@ -99,120 +100,127 @@ ActivityBase {
...
-                            property bool iAmSelected: modelData == items.colorSelector
-
-                            states: [
-                                State {
-                                    name: "notclicked"
-                                    when: !img.iAmSelected && !mouseArea.containsMouse
-                                    PropertyChanges {
-                                        target: img
-                                        scale: 0.8
+            Flickable{
```
petitlapin: add a space between the type and the brace (Flickable {)

petitlapin:
commited in master in https://quickgit.kde.org/?p=gcompris.git&a=commit&h=055426256882314feeec95d9337ae05fe846608c
Thank you

PR #160, opened 2016-10-08
Bargame port completed

iamutkarshtiwari:
Here is the ported QtQuick version of the Bargame activity (originally written in GTK+).

Link -https://phabricator.kde.org/T1538


src/activities/bargame/bargame.js
```diff
@@ -0,0 +1,214 @@
...
+
+function previousLevel() {
+    level--;
+    if (level < 1) {
+        level = 4;
+    }
+    currentLevel = level;
+    sublevel = 1;
+
+    reSetup(true);
```
iamutkarshtiwari: Forgot to remove. Made some logical changes to the function.

src/activities/bargame/bargame.js
```diff
@@ -0,0 +1,214 @@
...
+var tuxPositionFactor = [1.05, 1.8, 1.05]
+var ballSizeFactor = [0, 4, 4]
+var elementSizeFactor = [0, 4, 7]
+var noOfLables = [2, 3, 5]
+var moveCount = -1
+var level = 1
+var maxlevel = 4
+var sublevel = 1
+var numberOfSublevel = 3
+var bool = false
```
iamutkarshtiwari: It's unused. sorry

src/activities/bargame/bargame.js
```diff
@@ -0,0 +1,214 @@
...
+var boardPaused = 0
+var numberBalls = [[1, 4], [2, 6], [3, 6]]
+var boardSize = [15, 19, 29]
+var sampleBallsNo = [4, 6, 6]
+var tuxPositionFactor = [1.05, 1.8, 1.05]
+var ballSizeFactor = [0, 4, 4]
+var elementSizeFactor = [0, 4, 7]
+var noOfLables = [2, 3, 5]
+var moveCount = -1
+var level = 1
```
iamutkarshtiwari: level ~ currentlevel
maxLevel ~ maximum possible level

iamutkarshtiwari:
@petitlapin We can make this game playable in forced-landscape mode in Android phones. I don't know how to add it.

src/activities/bargame/Bargame.qml
```diff
@@ -0,0 +1,341 @@
...
+    onStart: focus = true
+    onStop: {}
+
+    pageComponent: Item {
+        id: rootWindow
+        signal start
+        signal stop
+        anchors.fill: parent
+
+        Image {
```
petitlapin: you can directly make this image the pageComponent

src/activities/bargame/Bargame.qml
```diff
@@ -0,0 +1,341 @@
...
+            x: rootWindow.width - rootWindow.width / 1.05
+            radius: 10
+
+            Image {
+                id: tux
+                source: Activity.url + "tux" + Activity.level + ".svg"
+                height: parent.height
+                width: parent.width
+                anchors.centerIn: tuxRect.Center
+                MouseArea {
```
petitlapin: I would just leave the Image without the hovering rectangle outside. It won't be visible on tablets and is not really useful/pretty (in my opinion :))

src/activities/bargame/Bargame.qml
```diff
@@ -0,0 +1,341 @@
...
+                    x: rootWindow.width / 2000
+                    y: rootWindow.height / 1.4
+                    columns: 15
+                    rows: 1
+                    Repeater {
+                        model: answerBallsPlacement.columns
+                        Image {
+                            id: greenBall
+                            source: Activity.url + "green_ball.svg"
+                            height: rootWindow.height / (9 + (Activity.sublevel - 1) * Activity.elementSizeFactor[Activity.sublevel - 1])
```
petitlapin: if all the balls and mask (should always) have the same width/height, you should create a readonly variable in the qml to assign it and use it in all the code.

src/activities/bargame/Bargame.qml
```diff
@@ -0,0 +1,341 @@
...
+                        }
+                    }
+                }
+            }
+
+            // Masks
+            Grid {
+                id: masks
+                x: rootWindow.width / 2000
+                y: rootWindow.height / 1.4
```
petitlapin: you can probably use anchors instead of x/y.
Can it be "merged" with the other grid?

src/activities/bargame/bargame.js
```diff
@@ -0,0 +1,212 @@
...
+.import QtQuick 2.0 as Quick
+
+var boardPaused = 0
+var numberBalls = [[1, 4], [2, 6], [3, 6]]
+var boardSize = [15, 19, 29]
+var sampleBallsNo = [4, 6, 6]
+var tuxPositionFactor = [1.05, 1.8, 1.05]
+var ballSizeFactor = [0, 4, 4]
+var elementSizeFactor = [0, 4, 7]
+var noOfLables = [2, 3, 5]
```
petitlapin: unused?

src/activities/bargame/bargame.js
petitlapin: What is the difference between level and currentLevel?
And between maxlevel and numberOfLevel (if I'm not wrong you removed numberOfLevel in the PR update)?
You should keep numberOfLevel to be consistent with the other activities (and respect the camelCase coding style).

src/activities/bargame/bargame.js
```diff
@@ -0,0 +1,212 @@
...
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+.pragma library
+.import GCompris 1.0 as GCompris
+.import QtQuick 2.0 as Quick
+
+var boardPaused = 0
```
petitlapin: unused (should be a boolean regarding the name if used)?

src/activities/bargame/bargame.js
```diff
@@ -0,0 +1,212 @@
...
+var numberBalls = [[1, 4], [2, 6], [3, 6]]
+var boardSize = [15, 19, 29]
+var sampleBallsNo = [4, 6, 6]
```
petitlapin: no need to abreviate names

src/activities/bargame/bargame.js
```diff
@@ -0,0 +1,212 @@
...
+var moveCount = -1
+var level = 1
+var maxlevel = 4
+var sublevel = 1
+var numberOfSublevel = 3
+var listWin = false
+var currentLevel = 0
+var items
+var noOfBalls = 1
+var tuxActive = true;
```
petitlapin: unused?

src/activities/bargame/bargame.js
```diff
@@ -0,0 +1,212 @@
...
+        level = 4;
+    }
+    currentLevel = level;
+    sublevel = 1;
+
+    reSetup();
+    initLevel();
+}
+
+function calculateWinPlaces() {
```
petitlapin: you should add a comment explaining your algorithm.

src/activities/bargame/bargame.js
```diff
@@ -0,0 +1,212 @@
...
+    if (playable.length != 0) {
+        var value = playable[Math.floor(Math.random()*playable.length)];
+    } else {
+        var value = randomNumber(numberBalls[sublevel - 1][0], numberBalls[sublevel - 1][1]);
+    }
+
+    play(false, value);
+}
+
+function play(human, value) {
```
petitlapin: instead of saying human/not human, I would put a number (player 1 or 2) and also a mode to know if it is a 2 players game or a one player (like in tictactoe or align4 for example). This way, we can add and handle a 2 players mode for this activity.

src/activities/bargame/bargame.js
```diff
@@ -0,0 +1,212 @@
...
+        machinePlay();
+    }
+}
+
+// Refreshes the scene
+function reSetup() {
+    items.tuxArea.enabled = true;
+    items.tuxArea.hoverEnabled = true;
+    moveCount = -1;
+    items.numberOfBalls.text = "1";
```
petitlapin: if you finish first sublevel or second, the minimum is not 1 but 2 or 3.

src/activities/bargame/bargame.js
```diff
@@ -0,0 +1,212 @@
...
+    noOfBalls = 1;
+
+    initLevel();
+
+    // Tux refresh
+    items.tux.source = url + "tux" + level + ".svg";
```
petitlapin: you don't need these 3 lines, it should be done automatically thanks to the binding

src/activities/bargame/bargame.js
```diff
@@ -0,0 +1,212 @@
...
+    items.tuxRect.y = items.rootWindow.height - items.rootWindow.height / 1.8;
+    items.tuxRect.x = items.rootWindow.width - items.rootWindow.width / tuxPositionFactor[sublevel - 1];
+
+    // Blue sample balls refresh
+    items.blueBalls.columns = sampleBallsNo[sublevel - 1];
```
petitlapin: you should use a binding for it directly in the qml file

src/activities/bargame/Bargame.qml
```diff
@@ -0,0 +1,341 @@
...
+            property alias backgroundImage: backgroundImage
+            property alias answerBallsPlacement: answerBallsPlacement
+            property alias answerBalls: answerBalls
+            property alias bar: bar
+            property alias bonus: bonus
+        }
+
+        onStart: {
+            Activity.start(items)
+            Activity.calculateWinPlaces();
```
petitlapin: it will be done only once at the beginning, is it what you expect or should it be done at every win/lose?

src/activities/bargame/bargame.js
iamutkarshtiwari: No, the minimum number of balls (according to the GTK+ version) at the start of a level should be 1 by default. This would change to the 'minimum number' for that level in every next moves.

It's like giving the kid a first chance to start with a small move to understand tux's strategy.

src/activities/bargame/Bargame.qml
iamutkarshtiwari: A readonly variable would create a problem with the dynamic size change of the elements during the gameplay which we might not want to happen. I'd rather try to shorten it a bit

src/activities/bargame/bargame.js
iamutkarshtiwari: The refresh needs to be done this way only. After having testes the binding technique, it seems that it doesn't refreshes the source dynamically during runtime. :/

src/activities/bargame/bargame.js
iamutkarshtiwari: The refresh needs to be done this way only. After having testes the binding technique, it seems that it doesn't refreshes the source dynamically during runtime. :/

iamutkarshtiwari: The only way to dynamize the content is through the Javascript code. ;)

iamutkarshtiwari:
@petitlapin Please have a look at the new commit and review my comments :)

src/activities/bargame/bargame.js
```diff
@@ -0,0 +1,209 @@
...
+}
+
+function previousLevel() {
+    level--;
+    if (level < 1) {
+        level = 4;
+    }
+    sublevel = 1;
+
+    reSetup();
```
petitlapin: I would put all the code of reSetup in initLevel (to keep the name consistency with all the other activities) if it has no side effect (the only difference I see is that except at start up we always call both in a row and reSetup calls initLevel)

src/activities/bargame/bargame.js
petitlapin: For me it's more a bug in Gtk+ version in this case. If the minimum for level 2 is 2, then at first play, the user shouldn't be able to play 1.

src/activities/bargame/bargame.js
iamutkarshtiwari: In this activity, a lot of variable inside the js file are interdependent. Moving these variables to the qml file would need a lot of changes throughout the complete code which might introduce unprecedent bugs.

iamutkarshtiwari:
Hi @petitlapin,
I have added the configuration dialog where user can set the difficulty level.
I haven't disturbed the data value inside js file because moving them to qml file caused some anomalous behaviour which caused acitivity crash. I haved tried to clean up the code a bit. 

P.S.- Trying to fix the bindings required moving the data values to qml file which disturbs the interdependency of variable which might introduce unnecessary bugs

iamutkarshtiwari:
Shall I move all these variable -
```
var numberBalls = [[1, 4], [2, 6], [3, 6]]
var boardSize = [15, 19, 29]
var sampleBallsNumber = [4, 6, 6]
var tuxPositionFactor = [1.05, 1.4, 1.05]
var ballSizeFactor = [0, 4, 4]
var elementSizeFactor = [0, 4, 7]
var moveCount = -1
var level = 1
var maxlevel = 4
var sublevel = 1
var listWin = []
var items
var numberOfBalls = 1

var url= "qrc:/gcompris/src/activities/bargame/resource/";
```
from JS file to QML file in order to make all the bindings work?

src/activities/bargame/Bargame.qml
```diff
@@ -0,0 +1,388 @@
...
+
+ActivityBase {
+    id: activity
+
+    onStart: focus = true
+    onStop: {}
+
+    pageComponent: Image {
+        id: rootWindow
+        source: "qrc:/gcompris/src/activities/bargame/resource/school_bg1.svg"
```
petitlapin: why don't use you the bar.level here as I said in the previous review? It would avoid having to change the background manually at each level change. You can keep also the Activity.url that would avoid to change every image paths if we need to change it for some reason

src/activities/bargame/Bargame.qml
```diff
@@ -0,0 +1,388 @@
...
+        onStart: {
+            Activity.start(items)
+            Activity.calculateWinPlaces();
+        }
+        onStop: { Activity.stop() }
+
+        // Tux image
+        Image {
+            id: tux
+            source: "qrc:/gcompris/src/activities/bargame/resource/tux1.svg"
```
petitlapin: same as for the background image

src/activities/bargame/Bargame.qml
```diff
@@ -0,0 +1,388 @@
...
+        }
+
+        // OK BUTTON
+        Image {
+            id: playLabel
+            x: rootWindow.width / 1.2
+            y: rootWindow.height / 2.4
+            width: rootWindow.width / 20
+            height: width
+            source: "qrc:/gcompris/src/activities/bargame/resource/ok.svg"
```
petitlapin: this ok button does not render well, you can use the one in core (bar_ok.svg) for now (I've created a OKButton.qml but not yet pushed it so we'll updated it later if needed)

src/activities/bargame/Bargame.qml
```diff
@@ -0,0 +1,388 @@
...
+            property alias greenBalls: greenBalls
+            property alias tuxArea: tuxArea
+            property alias boxes: boxes
+            property alias masks: masks
+            property alias numberLabel: numberLabel
+            property alias answerBallsPlacement: answerBallsPlacement
+            property alias answerBalls: answerBalls
+            property alias bar: bar
+            property alias bonus: bonus
+        }
```
petitlapin: I would add mode here and use it in replacement of Activity.sublevel (the name corresponds more to what is the value and you'll get the advantages of qml bindings)

src/activities/bargame/bargame.js
```diff
@@ -0,0 +1,196 @@
...
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+.pragma library
+.import GCompris 1.0 as GCompris
+.import QtQuick 2.0 as Quick
+
+var numberBalls = [[1, 4], [2, 6], [3, 6]]
```
petitlapin: For me, numberBalls and boardSize are related to the mode so I would have in the qml something like:
property var modes = [ "easy": {"minBalls": 1, "maxBalls": 4, "boardSize": 15 }, "medium": ...  ] (not sure if syntaxicly correct but the aim would be to have everything related to one mode in the same array).
If it is the same for the others, then you can do the same

src/activities/bargame/bargame.js
```diff
@@ -0,0 +1,196 @@
...
+                items.answerBallsPlacement.children[moveCount].opacity = 1.0;
+                items.answerBallsPlacement.children[moveCount].source = url + "green_ball.svg";
+            } else {
+                items.answerBallsPlacement.children[moveCount].opacity = 1.0;
+                items.answerBallsPlacement.children[moveCount].source = url + "blue_ball.svg";
+            }
+        }
+        if (moveCount == (boardSize[sublevel - 1] - 1)) {
+            // Level increment
+            level++;
```
petitlapin: I would not change level when winning, in this kind of games, the child may want to try multiple time as same level to find new ways of playing

iamutkarshtiwari:
Fixes-
1) Bindings fixed
2) scores added (top-right corner)
3) Ok svg changed
4) Level doesn't change on completion
5) configuration added to change levels

iamutkarshtiwari:
Fixes-
1) Bindings fixed
2) scores added (top-right corner)
3) Ok svg changed
4) Level doesn't change on completion
5) configuration added to change levels

src/activities/bargame/Bargame.qml
```diff
@@ -0,0 +1,395 @@
...
+                }
+            }
+        }
+
+        // Number of balls to be placed
+        Image {
+            id: ballNumberPlate
+            x: rootWindow.width / 1.2
+            y: rootWindow.height / 1.2
+            source: Activity.url + "enumerate_answer.svg"
```
petitlapin: Maybe we can use only one button for it and don't change the source, I don't see the interest of having it green at first then yellow. It's never resetted to green anyway

src/activities/bargame/Bargame.qml
```diff
@@ -0,0 +1,395 @@
...
+            id: bonus
+            Component.onCompleted: {
+                win.connect(Activity.initLevel)
+                loose.connect(Activity.initLevel)
+            }
+        }
+
+        Score {
+            id: score
+            message: "Score: " + items.score
```
petitlapin: use qsTr(). But for this, I would use 2 separate numbers, one for your score, the other for the opponent instead of having only one number that is incremented on win an decremented on lose.

stefant29:
Hi! While testing Bargame, i noticed some points you could improve:

1. User's "balls" appear under the bar buttons
2. Help menu: "prerequisite: brain" seems wrong: there should be something like "ability to count" or something more explicit.
3. Add a description in the "Help Menu" for the button that chooses how many "balls" the user wants to place: present how it works (e.g. "you have to press it to change the number of balls you want to place")
4. After the user's move, there should be a little delay, showing that the computer "thinks" for a second. Now Tux's answer is immediate and it doesn't give an overall good look .
5. In portrait mode, or when resizing the width of the window to a minimum, the "count button" goes behind the bar buttons. In this view, the "OK" button is very small and the image loses its aspect ratio.
6. The "balls" have a gradient that differentiate them from the background. Removing the gradient in inkscape or any other photo editor app would be an easy solution.

Apart from that, you did a good job on the visuals (haven't had time to check the implementation (the code). Keep it up!

Stefan

iamutkarshtiwari:
@stefant29 @petitlapin Thank you for the reviews :) I'll soon modify the patch based on the above specified requirements and will update you accordingly.

iamutkarshtiwari:
@petitlapin Please take a look on the newer commits and try the 2 player mode.

src/activities/bargame/ActivityInfo.qml
```diff
@@ -0,0 +1,35 @@
...
+
+ActivityInfo {
+  name: "bargame/Bargame.qml"
+  difficulty: 1
+  icon: "bargame/bargame.svg"
+  author: "Utkarsh Tiwari &lt;iamutkarshtiwari@kde.org&gt;"
+  demo: true
+  title: qsTr("Bargame (against Tux)")
+  description: ""
+  //intro: "put here in comment the text for the intro voice"
```
petitlapin: can you put here the good text (check in http://gcompris.net/wiki/Voices_translation to have the one needed)?

src/activities/bargame/bargame.js
petitlapin: comment stil valid, you should have only one table with all the data for each level instead of having multiple tables

src/activities/bargame/bargame.js
```diff
@@ -0,0 +1,212 @@
...
+    } else {
+        initiatePlayer2();
+        items.isPlayer1Turn = false;
+    }
+
+    calculateWinPlaces();
+    items.tuxArea.enabled = true;
+    items.tuxArea.hoverEnabled = true;
+    moveCount = -1;
+    items.numberLabel.text = "1";
```
petitlapin: it should not be 1 but the minimum for the selected difficulty.

src/activities/bargame/Bargame.qml
```diff
@@ -0,0 +1,666 @@
...
+
+ActivityBase {
+    id: activity
+    property int gameMode: 1
+    onStart: focus = true
+    onStop: {}
+
+    pageComponent: Image {
+        id: rootWindow
+        source: Activity.url + "school_bg" + bar.level + ".svg"
```
petitlapin: There is an error at beginning (bar.level is still 0), can you handle it? Same for qrc:/gcompris/src/activities/bargame/Bargame.qml:96:9: QML Image: Cannot open: qrc:/gcompris/src/activities/bargame/resource/tux0.svg

iamutkarshtiwari:
@petitlapin  I have fixed most of the issues mentioned above. Major once -
- prerequisite updated
- description updated
- delay for Tux's move added
- count Button now moves above bar in portrait mode and ok button size increased
- terminal errors resolved

Moving all the array based data into a single array will create a lot of confusion. And that's something we should avoid :)

iamutkarshtiwari:
@petitlapin Fixed the aspect ratios of the image to look better in portrait mode as well. Re-positioned the game elements for better space utilization.

@stefant29 Could you please help me remove gradients from the images? I am not much familiar with inkscape.

rahulyadav170923:
@iamutkarshtiwari 
1. the pointer for increasing no. of balls is faulty for medium and hard mode .
2. for the pointer there should be increase and decrease arrows (personal opinion)(check with your mentors also)
3.you shouldn't allow more balls to be put when there less space even at the end. this might create confusion.

rahulyadav170923:
@iamutkarshtiwari  1 .you have chosen a cross and blue circle to mark the players. I think it  should be the the colours of the balls being used ( ask you mentors also)
2. Put some space btw Ok and the ball counter .
3. arrange the ball and counter inside the rectangle.

iamutkarshtiwari:
@petitlapin Please test the latest commit both on Android (apk) and Desktop

petitlapin:
I created a patch to apply to last version: https://drive.google.com/open?id=0B1Wylu9YXKcKNVJwendTUXJqWUk 
I started it before your last commits so maybe there are some differences. We can discuss about it, you don't have to accept everything blindly :).

For the images/style, don't bother too much, we'll see with Timothée once merged.
I agree with the cross/circle being changed to the good color but they'll probably change (they are too close for now) so I don't think it's necessary to update them now (except if you want to play a little with inkscape of course).
So globally, I'm satisfied with it :). There is one last point that I'd like to try: in vertical mode (on phone), it's quite difficult to see the balls on hard mode (29 to fill on a small place). Can you test to have the bar vertically and see if it renders well or if we keep like now?

iamutkarshtiwari:
@petitlapin Portrait mode is now improved. Please test it on an Android device and let me know your suggestions.

stefant29:
hi! did you solve your "gradieng" problems?

iamutkarshtiwari:
@stefant29 I didn't, actually couldn't :/ Could you please do it for me? I can share the two svgs for the blue and green balls.

stefant29:
send me a private message on irc (StefanT29_) and we'll see if i can do something about it

iamutkarshtiwari:
@petitlapin Manual merge done 👍

iamutkarshtiwari:
@Animtim Please review this patch

petitlapin:
Hi, I merged it in master. I did other changes on your code, please check them.

Thank you for the activities,

Johnny

PR #164, opened 2016-11-01
Rahulyadav summer squash

rahulyadav170923:
Created Guesscount Activity


src/activities/guesscount/DragTile.qml
```diff
@@ -0,0 +1,52 @@
...
+            id: tile
+        }
+    }
+    Component.onCompleted: {
+        if(type == "operators")
+            createDynamicObject()
+    }
+    onChildrenChanged: {
+        if(type == "operators"){
+            if(root.children.length == 0)
```
petitlapin: I'm not sure to understand the 0 and 3 there.
For 0, I understand that we need to create a new tile to replace the one we moved. But why 3, I think there might be a potential memory leak there. If you only drop one tile, it will allocate a new Tile but if you change level, it's never freed.

src/activities/guesscount/Tile.qml
```diff
@@ -0,0 +1,80 @@
...
+            fontSize: mediumSize
+        }
+        states: [
+            State {
+                when: mouseArea.drag.active
+                ParentChange { target: tile; parent: root }
+                AnchorChanges { target: tile; anchors.verticalCenter: undefined; anchors.horizontalCenter: undefined }
+            },
+            State {
+                when: dialogVisible
```
petitlapin: it should be items.dialog.visible. Using so, you don't need onPressed, onClicked and changeDialogVisibility function.
You can also bind it directly without a state.

src/activities/guesscount/Guesscount.qml
```diff
@@ -0,0 +1,415 @@
...
+            property Item main: activity.main
+            property alias background: background
+            property alias bar: bar
+            property alias bonus: bonus
+            property int sublevel: 0
+            property alias operandRow: operandRow
+            property  var data
+            property int result: data[sublevel-1][1]
+            property alias timer: timer
+            property alias dialog: dialog
```
petitlapin: can you rename the variable. For example, warningDialog would be more precise.

src/activities/guesscount/Guesscount.qml
```diff
@@ -0,0 +1,415 @@
...
+                                    levelOperators: items.levelArr
+                                    width: background.width
+                                    height: background.height/10
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            onClose: {
```
petitlapin: instead of redefining the onClose function, you can use the dataValidationFunc attribute. You can take a look at  balancebox where it's used.

rahulyadav170923:
thanks for the review johnny. i'll make the changes

petitlapin:
Merged on kde git directly, thank you!

PR #171, opened 2017-01-03
Hangman attempted chars

RudraNilBasu:
Displays a list of characters attempted by the user.
* checked with resizing the screen and enabling the image.
* qsTr() is used for strings that are to be translated
* All types of characters will be included in the list, and not just A-Z English characters.
(Since whatever string comes to the processKeyPress(text) function as parameter is 
added to the list).

### DIFF
--- a/src/activities/hangman/hangman.js
+++ b/src/activities/hangman/hangman.js
@@ -71,11 +71,28 @@ function start(items_) {
 function stop() {
 }
 
+function initGuessText() {
+
+    items.guessedText.text="";
+    var c=0;
+    for(var i=65;i<=90;i++) {
+        c++;
+        items.guessedText.text=items.guessedText.text.concat(String.fromCharCode(i).toUpperCase());
+        if(c==8) {
+            c=0;
+            items.guessedText.text=items.guessedText.text.concat('\n');
+        } else {
+            items.guessedText.text=items.guessedText.text.concat('\t');
+        }
+    }
+}
+
 function initLevel() {
     items.bar.level = currentLevel + 1;
     var currentLesson = lessons[currentLevel];
     wordList = Lang.getLessonWords(dataset, currentLesson);
     Core.shuffle(wordList);
+    initGuessText();
 
     maxSubLevel = wordList.length;
     items.score.numberOfSubLevels = maxSubLevel;
@@ -135,6 +152,9 @@ function processKeyPress(text) {
     }
     // Add the character to the already typed characters
     alreadyTypedLetters.push(text);
+
+    // Remove the letter from the guessed Text
+    items.guessedText.text=items.guessedText.text.replace(text.toLocaleUpperCase(), " ");
 
     // Get all the indices of this letter in the word
     var indices = [];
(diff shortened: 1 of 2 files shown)


src/activities/hangman/hangman.js
```diff
@@ -76,6 +76,7 @@ function initLevel() {
     var currentLesson = lessons[currentLevel];
     wordList = Lang.getLessonWords(dataset, currentLesson);
     Core.shuffle(wordList);
+    items.guessedText.text=qsTr("Attempted: ");
```
petitlapin: why is it initialized twice?

src/activities/hangman/hangman.js
```diff
@@ -136,6 +137,9 @@ function processKeyPress(text) {
     // Add the character to the already typed characters
     alreadyTypedLetters.push(text);
 
+    // add
+    items.guessedText.text=items.guessedText.text.concat(text, ", ");
```
petitlapin: it's quite tricky but I would keep all the letters and rebuild the typed letters string using:
items.guessedText.text = qsTr("Attempted: %1").arg(usedLetters)
usedLetters can probably be built using alreadyTypedLetters and join (something like qsTr("Attempted: %1").arg(alreadyTypedLetters.join(", ")).
For translation purpose, it's better to have the whole context (we could have a language that would need to display it on the other direction: "used letters: Attempted" and if we do like now, we won't be able to do it as the code supposes that the letters should be displayed on right of "Attempted: ")

RudraNilBasu: In the recent commit, I solved it in a similar way in the _form()_ method

In that method, I initialized a string with "Attempted: ". Then I sorted the "alreadyTypedLetters" 
array and inserted it into the forming string one by one. When the full traversal is over, I am 
returning the string after passing it through qsTr(), and that is shown in the top of the activity.

RudraNilBasu:
@petitlapin thank you for the review. I have updated the PR. These problems should be fixed 
* the warning related to the anchors is sorted out
* the layout is fixed, the text doesn't appear on top of the hints image. The text is placed on a Rectangle,
to enhance the visibility
* Characters in the text are sorted
* Fixed translation issues in a language that displays in the other direction: "used letters: Attempted"

src/activities/hangman/hangman.js
```diff
@@ -136,6 +136,9 @@ function processKeyPress(text) {
     // Add the character to the already typed characters
     alreadyTypedLetters.push(text);
 
+    // add
+    items.guessedText.text=form()
```
petitlapin: form is not a good name for a function, maybe you can call it more explicitely.
Also, can you add a space before and after the '=' sign?

src/activities/hangman/hangman.js
```diff
@@ -200,6 +203,8 @@ function initSubLevel() {
     alreadyTypedLetters = new Array();
     currentWord = items.goodWord.translatedTxt;
     items.hidden.text = ""
+    items.guessedText.text=qsTr("Attempted: ");
```
petitlapin: I would call form() here, and in form() instead of returning the string, I would directly update items.guessedText.text like stated below

RudraNilBasu:
@petitlapin Thank you for the review, I am away from keyboard for few days, will make the changes when I get back with it.

RudraNilBasu:
@petitlapin updated the PR, the issues are solved in this commit.

petitlapin:
merged in kde master branch, thank you!

PR #172, opened 2017-01-04
Ascending order activity

RudraNilBasu:
Ascending order activity

Given few randomly generated numbers, user will have to click on the numbers in 
ascending order.


src/activities/ascending_order/ascending_order.js
```diff
@@ -0,0 +1,155 @@
...
+            items.bonus.bad("lion")
+            return
+        }
+    }
+    // CORRECT answer
+    items.bonus.good("lion")
+}
+
+function printSorted() {
+    for(var i=0;i<entered.length-1;i++) {
```
petitlapin: I think there is a sort function in js directly

src/activities/ascending_order/ascending_order.js
```diff
@@ -0,0 +1,155 @@
...
+        for(var j=0;j<entered.length-i-1;j++) {
+            if(Number(entered[j]>Number(entered[j+1]))) {
+                var temp=entered[j+1];
+                entered[j+1]=entered[j]
+                entered[j]=temp
+            }
+        }
+    }
+    for(i=0;i<entered.length;i++) {
```
petitlapin: you can use Array.join()

allon49:
Hi Rudra,

I had a look to your number ordering activity.
For me it is a very good idea.
Ordering activity is one of the most efficient activity to check if a kid has understood the decimal system (121 is bigger than 12 for example).
I tryed your activity and would like to suggest some improvments.

1- It should have a ok button for the kid to confirm when he is ready to validate.
2- Johnny found an application on the web which is exactly what I would like to see on gcompris. You can reproduce the behaviour using flow:

http://doc.qt.io/qt-4.8/qml-flow.html

3- There should be options like in the web application to match to different ages. You can have a look to divyam categorisation to see how he manages settings.

Good luck,

Emmanuel

RudraNilBasu:
@petitlapin @allon49 In the ascending order activity: 
1. The items are moved by drag and drop to reposition the items
2. The tiles are now shifted for repositioning instead of swapping the positions
3. There is an ok button to allow the user to confirm when ready
4. There are 4 levels each with 3 sub levels, the sub levels are of the same difficulty (difficulty in the form
     of number of tiles and the range of numbers)

src/activities/ascending_order/ActivityInfo.qml
```diff
@@ -0,0 +1,35 @@
...
+ * along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+import GCompris 1.0
+
+ActivityInfo {
+  name: "ascending_order/Ascending_order.qml"
+  difficulty: 2
+  icon: "ascending_order/ascending_order.svg"
+  author: "Rudra Nil Basu &lt;rudra.nil.basu.1996@gmail.com&gt;"
+  demo: false
```
petitlapin: this value should be true

src/activities/ascending_order/ActivityInfo.qml
```diff
@@ -0,0 +1,35 @@
...
+  title: qsTr("Ascending Order")
+  description: qsTr("Arrange the given numbers in ascending order")
+  //intro: "click on the numbers in ascending order"
```
petitlapin: can you update this. It corresponds to the intro voice that should be listened to explain the activity.

src/activities/ascending_order/ActivityInfo.qml
```diff
@@ -0,0 +1,35 @@
...
+  goal: qsTr("arranging numbers in ascending order")
+  prerequisite: qsTr("Move, drag and drop using mouse")
+  manual: qsTr("You are provided with few numbers. Drag and Drop the numbers in it's correct position to reorder the numbers in ascending order")
+  credit: "openclipart.org"
```
petitlapin: I'm not sure it is needed, we used the images in a lot of places.

src/activities/ascending_order/Ascending_order.qml
```diff
@@ -0,0 +1,218 @@
...
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+import QtQuick 2.1
+import GCompris 1.0
+
+import "../../core"
+import "ascending_order.js" as Activity
+import "qrc:/gcompris/src/core/core.js" as Core
```
petitlapin: not needed

petitlapin:
Hi,
globally it's good (I have to take a closer look at the code).
Can you add the sublevel score (Score item in the core folder)?
Also, click on next/previous should change the level (and reset the current sublevel).
I would put the ok button above the bar too, else when resizing, they override

Thank you,

Johnny

divyam3897:
hi,
I faced one more issue while playing it, for the extreme numbers (corner ones) if I drag a number beyond them then number came back and it was placed at the required position only if a portion of it overlapped with the corner number. Shouldn't the number be placed ahead of it (at its desired position) in that case?

Thank you
Divyam Madaan

RudraNilBasu:
@divyam3897 hi, the numbers will change the position (and the others will shift back or front) only when one tile is placed over another tile (not *exactly* over, a minimum distance is maintained). In that case, the dragged number will occupy the position of the tile on which it was dragged and the rest of them will get shifted accordingly. So, if a tile is dragged to the extreme corner, it crosses the minimum distance and hence it comes back to the original position.

divyam3897:
I did a screenshot for some cases. 
1. http://pasteboard.co/EehVhPRf2.png In this I was able to produce a gap between the numbers which should not be the case? (was not always reproducible but occured at times)
2. http://pasteboard.co/EejDEXNZv.png In this the tile (29 the one being dragged) is pretty close and maybe should not go back?

RudraNilBasu:
@divyam3897 
1. I didn't get the first case. Do the tiles remain in that way (separated by a space) after dropping a tile (after the mouse is released)? In that case is there any specific way by which it can be reproduced? Also, which number is being dragged in this case? Strange. This type of thing never actually occurred to me.

2. This one is okay. The 29 tile should be placed over the 74 tile in order to change the positions. The minimum distance is less in this case because had there been another tile on the left of 74, then the 29 tile should have occupied that position. The aim is to place a tile over the tile where it should belong.

allon49:
Hi,
Divyam, did you do the screenshot while dragging?
Emmanuel


2017-03-01 17:55 GMT+01:00 Rudra Nil Basu <notifications@github.com>:

RudraNilBasu:
@petitlapin Following changes were made:
1. demo is set to true
2. updated intro
3. Removed credit and unused imports
4. Added sub level score
5. Click on next/previous changes the level, while updating current sub level to zero
6. Repositioned ok button so that it doesn't get overridden by the bar.

divyam3897:
@allon49 yes, In one case the card returned back after the screenshot and in the other it stayed there, creating an empty place.
@RudraNilBasu The sublevel count also may need to be repositioned? It gets hidden in case of small screen size http://pasteboard.co/Ey6QDoCx7.png

RudraNilBasu:
Repositioned the sub level score to avoid getting hidden
under the bar for small screen sizes.

The card coming back is normal, since it is not over any
other card and it is going outside the play area once
released, so the move is invalid and thus returns back
to the original position.

The card staying there is quite surprising. Does the
problem still persist ? Strange, I can't seem to
reproduce it in any way.

src/activities/ascending_order/ascending_order.js
```diff
@@ -0,0 +1,239 @@
...
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+.pragma library
+.import QtQuick 2.0 as Quick
+
+var currentLevel = 0
+var numberOfLevel = 4
+var currentSubLevel = 0
```
petitlapin: you can directly use the score for this instead of having 2 other variables

RudraNilBasu:
@petitlapin Following changes were made:
* `numberOfSublevels` and `currentSublevels` are used from `Score.qml` instead of using extra variables.
* Used `alias` in `score` in `QtObject`
* Number of sublevels is increased to 4
* Each sublevel contains 2 ascending and 2 descending order arrangements
* Added alphabetical order arrangement

RudraNilBasu:
@petitlapin I have changed the alphabets activity to support various languages other than english.
Also, for testing purposes, I have tested it by replacing (not shown in code)
```
var letters = qsTr("a/b/c/d..../z")
```

with 

```
var letters = qsTr("À/à/Â/â/Æ/æ/Ç/ç/È/è/É/é/Ê/ê/Ë/ë/Î/î/Ï/ï/Ô/ô/Œ/œ/Ù/ù/Û/û/Ü/ü/«/»/€/₣")
```

src/activities/alphabetical_order/Alphabetical_order.qml
```diff
@@ -0,0 +1,30 @@
+/* GCompris - alphabetical_order.qml
+ *
+ * Copyright (C) 2016 Rudra Nil Basu <rudra.nil.basu.1996@gmail.com>
```
petitlapin: 2017

src/activities/ascending_order/ascending_order.js
```diff
@@ -0,0 +1,287 @@
...
+    for(var i = 0;i < items.boxes.model;i++) {
+        if (mode == "number") {
+            items.boxes.itemAt(i).boxValue = num[i].toString()
+        } else if (mode == "alphabets") {
+            items.boxes.itemAt(i).boxValue = num[i]
+        }
+    }
+}
+
+function generateRandomNumbers() {
```
petitlapin: need to be renamed more generic

src/activities/alphabetical_order/Alphabetical_order.qml
```diff
@@ -0,0 +1,30 @@
...
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+import QtQuick 2.1
+
+import "../../core"
+import "../ascending_order"
+
+Ascending_order {
```
petitlapin: camelCase (also for filename). Same for AlphabeticalOrder

src/activities/ascending_order/ascending_order.js
```diff
@@ -0,0 +1,287 @@
...
+
+    noOfTilesInPreviousLevel = -1
+    initLevel()
+}
+
+function stop() {
+}
+
+function initLevel() {
+    ascendingOrder = items.score.currentSubLevel % 2 == 0 ? true : false
```
petitlapin: to be checked on the mailing list (or at least with Emmanuel) but I wouldn't change every sublevel. It would not be efficient to understand the meaning behind ascending/descending order

src/activities/ascending_order/ascending_order.js
```diff
@@ -0,0 +1,287 @@
...
+    mode = mode_
+    currentLevel = 0
+    items.flow.validMousePress = false
+    thresholdDistance = 4000 * items.ratio
+    items.score.currentSubLevel = 0
+    items.score.numberOfSubLevels = 4
+
+    if (mode == "alphabets") {
+        //: list containing all the characters separated by a "/"
+        var letters = qsTr("a/b/c/d/e/f/g/h/i/j/k/l/m/n/o/p/q/r/s/t/u/v/w/x/y/z")
```
petitlapin: I think having a help button showing the order could be nice.

petitlapin:
@allon49 if we do like that, we need to think of other activities where it could be applied to (not for this PR but it could be in the improvements section). With one generic submenu, if we manage to load datasets easily, then we can think of having the difficulty for each category instead at activity level (which would not make sense on this case as the activity level is more global than only a 2 years old span).
For the real words, it would need a new specific dataset so for now, I guess it could be a future improvement.
The question is to know if we do a first version like the current one (so each level corresponds to a range and we have sublevels in it) or do we directly do the menu one (but it would need to be generic enough to be "easily" adapted to the activities that need it)?

petitlapin:
Merged in ascending_order branch to continue the work

PR #173, opened 2017-01-10
Reduce SVG and PNG resource size

IlyaBizyaev:
All of the changes are losless.


IlyaBizyaev:
The following scripts were used to reduce image sizes:
```
from os import system
 
f = open("svg.txt")
files = f.readlines()
for item in files:
        system("inkscape -z --verb=FileVacuum ./compress/"+item)
f.close()
```

```
from os import system
 
f = open("png.txt")
files = f.readlines()
for item in files:
        system("optipng -o7 ./GCompris-qt/"+item)
f.close()
```

"svg.txt" and "png.txt" are text documents with lists of relative paths to images, produced with the _find_ shell command.
Inkscape and Optimage were installed from openSUSE repositories.

petitlapin:
I merged it directly in the kde master. Thank you!

PR #177, opened 2017-01-27
Add a mark where lines can't be drawn for activities "drawnumber" and "drawletter"

RudraNilBasu:
Previously a node was disabled whenever clicked. Adding an indicator to the last node will
make it easier to visualise from where the line is to be drawn, especially in the case of the very first node, where the node just disappears (http://pasteboard.co/r5PZTPZ9I.png and http://pasteboard.co/r5Y30iNiF.png ).


RudraNilBasu:
@petitlapin changes made since the last commit: 
* removed redpoint, used better marker for the point
* instead of marking the previously traversed node every time,
it marks the positions when a line cannot be drawn, ie:
  * clicking on the first node
  * between nodes where no line needs to be drawn

petitlapin:
applied directly in master. Thank you!

PR #178, opened 2017-02-02
Added Odd even category, changed dataset addition technique

RudraNilBasu:
* Added an odd-even category in the categorization activity

* changed the techniques used to add datasets in the activity. 
Originally the datasets were named category<xx>.qml. Now, the 
datasets can have any name, the FileNames::getFiles(String)
returns the list of files in the location, and using that, the datasets
are extracted, and using the isEmbedded variable, we decided 
whether the dataset is a demo or not.


src/activities/categorization/Categorization.qml
```diff
@@ -36,6 +36,7 @@ ActivityBase {
 
     property string boardsUrl: "qrc:/gcompris/src/activities/categorization/resource/"
     property bool vert: background.width < background.height
+    property string datasetURL: "../../GCompris-qt/src/activities/categorization/resource/board/"
```
petitlapin: you need to use the qrc path, not a relative path, like boardsUrl

src/activities/categorization/categorization.js
```diff
@@ -51,11 +51,18 @@ function init(items_,boardsUrl_) {
 function start() {
     categoriesData = []
     items.categoryReview.stop()
+
+    var s = items.categories.split(" ")
+    var flag = items.file.exists(fileName) ? 1 : 0
```
petitlapin: if it is a boolean, use true and false. And a real name for the variable

src/activities/categorization/categorization.js
```diff
@@ -51,11 +51,18 @@ function init(items_,boardsUrl_) {
 function start() {
     categoriesData = []
     items.categoryReview.stop()
+
+    var s = items.categories.split(" ")
```
petitlapin: name the variable correctly

src/core/FileNames.h
```diff
@@ -0,0 +1,37 @@
...
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef FILENAMES_H
+#define FILENAMES_H
+
+#include <dirent.h>
```
petitlapin: why?

src/core/FileNames.cpp
```diff
@@ -0,0 +1,52 @@
...
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "FileNames.h"
+
+#include <dirent.h>
```
petitlapin: why?

src/core/FileNames.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include <QString>
+#include <string.h>
```
petitlapin: why?

src/core/FileNames.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include <QObject>
+
+
+FileNames::FileNames()
+{
+}
+
+QString FileNames::getFiles(QString location)
```
petitlapin: use const &

src/core/FileNames.cpp
```diff
@@ -0,0 +1,52 @@
...
+{
+    const char *s = location.toUtf8().constData();
+    DIR *dpdf = opendir(s);
```
petitlapin: use QDir which is portable

src/core/FileNames.cpp
```diff
@@ -0,0 +1,52 @@
...
+    struct dirent *epdf;
+
+    QString files;
+
+    if(dpdf != NULL) {
+        while(epdf = readdir(dpdf)) {
+            if(strcmp(epdf->d_name, ".") != 0 && strcmp(epdf->d_name,"..")) {
+                // next file name = epdf->d_name
+                files.append(QString::fromLocal8Bit((epdf->d_name)));
```
petitlapin: use a QList/QStringList

src/core/FileNames.cpp
```diff
@@ -0,0 +1,52 @@
+/* GCompris - FileNames.cpp
```
petitlapin: it would be better to call it Dir or Directory

src/core/main.cpp
```diff
@@ -268,6 +272,9 @@ int main(int argc, char *argv[])
         window->show();
     }
 
+    QScopedPointer<FileNames> files(new FileNames);
```
petitlapin: It would be better to keep the same instanciation as File or other classes

src/core/FileNames.h
```diff
@@ -0,0 +1,37 @@
...
+#include <QString>
+#include <QObject>
+
+class FileNames : public QObject
```
petitlapin: Add a quick documentation to what the class do (you can check File for example)

src/activities/categorization/resource/board/category_odd_even.qml
```diff
@@ -23,38 +23,38 @@ import QtQuick 2.0
...
-            "image": imagesPrefix + "numbers.jpg",
+            "image": imagesPrefix + "eleven.png",
             "content": [
                 {
                     "instructions": qsTr("Place the EVEN Numbers to the right and ODD Numbers to the left"),
-                    "image": imagesPrefix + "even2.jpg",
-                    "maxNumberOfGood": 4,
+                    "image": imagesPrefix + "sixteen.png",
+                    "maxNumberOfGood": 5,
                     "maxNumberOfBad": 4,
```
divyam3897: As mentioned use more images in initial levels (teaching part) . Add 12 images in level1, 9 images in level2, 6 images in level 3 (concept building) , 6 images in level 4 (testing).

src/activities/categorization/resource/board/category_odd_even.qml
```diff
@@ -0,0 +1,63 @@
...
+                    "bad": ["odd47.svg","odd15.svg","odd35.svg","odd23.svg","odd39.svg","odd111.svg"]
+                },
+                {
+                    "instructions": qsTr("Place the EVEN Numbers to the right and ODD Numbers to the left"),
+                    "image": imagesPrefix + "two.png",
+                    "maxNumberOfGood": 4,
+                    "maxNumberOfBad": 5,
+                    "prefix": "qrc:/gcompris/src/activities/categorization/resource/images/odd_even/",
+                    "good": ["even786.svg","even276.svg","even192.svg","even92.svg"],
+                    "bad": ["odd111.svg","odd145.svg","odd39.svg","odd353.svg","odd23.svg"]
```
divyam3897: Did you create the images on your own? If not add where you took it from with the correct licence in README as done for other images.

src/activities/categorization/Categorization.qml
RudraNilBasu: The qrc path was not working with opendir(). It returned NULL on using the qrc path.

src/activities/categorization/resource/board/category_odd_even.qml
RudraNilBasu: @divyam3897 yes, all the images in qrc:/gcompris/src/activities/categorization/resource/images/odd_even/ are created by me.

I also used the images from word_sample in the first level.

src/activities/categorization/Categorization.qml
petitlapin: Using opendir is a bad choice here: it's not portable and it won't work once the software is installed. ../GCompris-qt works for you because the source code of gcompris is on this path/ Once installed, the source code is not installed so you no more have access to this path.

RudraNilBasu: I shifted from using opendir() to QDir, but it still doesn't work for the qrc path.
Code - http://pastebin.com/JavNCD4G

RudraNilBasu: Is there anything I am missing here ?

RudraNilBasu:
* Used QDir instead of opendir()
* Removed use of relative paths
* Renamed Filename.h/cpp to Directory.h/cpp
* Added documentation for Directory class
* getFiles(QString) now returns QStringList
* Changed instantiation technique
* No of levels in odd-even category is now 
increased to 6
@petitlapin @divyam3897 can you have a look at it?

src/core/main.cpp
```diff
@@ -268,6 +272,8 @@ int main(int argc, char *argv[])
         window->show();
     }
 
+    engine.rootContext()->setContextProperty("files", new Directory);
```
petitlapin: you don't need this (and the added include on top)

src/core/Directory.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+Directory::Directory()
+{
+}
+
+QStringList Directory::getFiles(const QString& location)
+{
+    QStringList fileNames;
+    QDir dir(location);
+    QFileInfoList files=dir.entryInfoList();
```
petitlapin: you can directly use http://doc.qt.io/qt-5/qdir.html#entryList

src/core/Directory.cpp
```diff
@@ -0,0 +1,52 @@
...
+            // if it is a file
+            fileNames.append(file.fileName());
+        }
+    }
+    return fileNames;
+}
+
+void Directory::init()
+{
+    qmlRegisterType<Directory>("GCompris", 1, 0, "directories");
```
petitlapin: keep the consistency with the other qmlRegisterType

src/activities/categorization/categorization.js
```diff
@@ -51,11 +51,19 @@ function init(items_,boardsUrl_) {
 function start() {
     categoriesData = []
     items.categoryReview.stop()
+
+    var categoryLists = []
+    categoryLists = items.categories
+    var allFilesExist = items.file.exists(fileName) ? true : false
```
petitlapin: isEmbeddedMode is better as name

RudraNilBasu:
@petitlapin Updated the PR, the following issues should now be fixed
* Removed the use of setContextProperty()
* Used entryList() instead of entryInfoList()
* Renamed allFilesExist -> isEmbeddedMode
* Optimised the size of the images used in
  odd-even category

petitlapin:
Thank you for the PR, I commited it in the main repo:
https://cgit.kde.org/gcompris.git/commit/?id=955462b943c34fc130d1a68fcfb0e1ec6393a3f0 (for the directory class)
https://cgit.kde.org/gcompris.git/commit/?id=db7a4a9b743a3521c7c68f0b2b54719cbc9582db (for the level).
I did a few changes, feel free to check them. Most notable is that I put the odd_even category in the "full" version. To test it, you can generate the words.rcc from https://github.com/gcompris/GCompris-words and put it in ~/.cache/KDE/gcompris-qt/data2/words/ and start again GCompris.

PR #180, opened 2017-02-25
Task T4798 : Improving Letter-in-word activity

soura97:
Instead of showing "bad" bonus, when a wrong word is selected, the kid can be asked to check answer. This results in the word not getting selected and the kid stays on the same sublevel.


petitlapin:
The work is in progress on phabricator (https://phabricator.kde.org/D4938), can we close the PR?

soura97:
Yes. We can close this now.

PR #181, opened 2017-03-07
categorization:expert mode only on some categories

RudraNilBasu:
The expert mode on the categorization activity chooses randomly
from all the available categories. But among them, there may be
few categories which doesn't need an expert mode (example: the odd
even category)

The indices of the categories which are loaded in the current level
and which allows the expert mode is stored in an array. While
choosing a random index during expert mode, it is checked if it is
present in the array, if not, the randomisation process is repeated

Signed-off-by: Rudra Nil Basu <rudra.nil.basu.1996@gmail.com>

### DIFF
--- a/src/activities/categorization/categorization.js
+++ b/src/activities/categorization/categorization.js
@@ -212,12 +212,9 @@ function getCategoryLevels() {
     }
     // If expert mode is selected, select a random level (selectedLevel) from a random category (selectedCategory)
     else if(items.mode === "expert") {
-        var selectedCategory = Math.floor(Math.random() * lessons.length)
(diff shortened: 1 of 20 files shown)


RudraNilBasu:
@petitlapin @divyam3897 can you have a look at this?

RudraNilBasu:
@petitlapin The selection of categories in the expert mode of categorization now does not need to use the
index. The `expertCategories` store the content of the levels which allows expert mode and while choosing
a random level, the levels are now selected from `expertCategories` array instead of the `lessons` array

petitlapin:
thanks, merged on https://commits.kde.org/gcompris/cc25f147d53bb3f708f2464a5052ec4faca26782

PR #182, opened 2017-03-19
lang, add Polish dataset

wojnilowicz:
Please add me to translators list if applicable.


petitlapin:
Hi,

I can't commit the file in the kde repository due to the email address you use for github: Email address has an invalid domain : lukasz@luklap.workgroup
Which other one can I use?
There is no translators list but with this commit, your name will be in the developers list we display in the about panel.

Thank you,

Johnny

PR #184, opened 2017-03-30
Config_Scroll:Hide "scroll_down" at bottom of page

RudraNilBasu:
The scroll_down was visible even when we scrolled till the bottom
of the screen in the configuration menu. Since the page position
is in the range from 0.0 to 1.0 - size ratio, the range will be
from 0.0 to 1.0 - flick.visibleArea.heightRatio. Using this, the
scroll_down image gets disappeared when we are at the bottom of
the configuration page

Signed-off-by: Rudra Nil Basu <rudra.nil.basu.1996@gmail.com>

### DIFF
--- a/src/core/DialogActivityConfig.qml
+++ b/src/core/DialogActivityConfig.qml
@@ -218,7 +218,7 @@ Rectangle {
                     onUp: flick.flick(0, 1400)
                     onDown: flick.flick(0, -1400)
                     upVisible: flick.visibleArea.yPosition <= 0 ? false : true
-                    downVisible: flick.visibleArea.yPosition >= 1 ? false : true
+                    downVisible: flick.visibleArea.yPosition >= 1.0 - flick.visibleArea.heightRatio ? false : true
                 }
             }
 


RudraNilBasu:
@Animtim @petitlapin can you have a look at this?

Animtim:
Hi, I answered on irc, but I copy a the most important parts:

-one thing first: please, don't use github for anything...Use phabricator for patch review or task planning, or the kde bug tracker for reports. Github is just a mirror for us, ideally nothing else should happen there.

-about the PR itself: thanks :) . Though, I already tried your suggested fix for the button, but it actually doesn't really work, as it disappears and then comes back as the scroll finishes (as I guess it goes back up slightly, or something..); I also tried a few other things, but ended leaving the bottom one static for now, until the bigger design update soon.

RudraNilBasu:
Adding few points from the discussions on IRC for future references

* The core problem is the scroll_down button is not going away on some devices. While it works well on
some devices, the button reappears after the scrollbar moves up slightly up after reaching the end.
* I am closing this PR since @Animtim is working on this (and further features)

PR #190, opened 2017-07-20
Categorization: Add another mode and audio at drop

amitsagtani97:
[Link to Task Details](https://phabricator.kde.org/D5891)

### DIFF
--- a/src/activities/categorization/Categorization.qml
+++ b/src/activities/categorization/Categorization.qml
@@ -122,6 +122,7 @@ ActivityBase {
                     height: dialogActivityConfig.height
                     property alias instructionsBox: instructionsBox
                     property alias scoreBox: scoreBox
+                    property alias categoryImageBox: categoryImageBox
 
                     GCDialogCheckBox {
                         id: instructionsBox
@@ -129,16 +130,9 @@ ActivityBase {
                         text: qsTr("Instructions visible")
                         checked: items.instructionsVisible
                         onCheckedChanged: {
-                            if(instructionsBox.checked) {
-                                items.instructionsVisible = true
-                                Activity.selectmode()
-                            }
-                            else if(!instructionsBox.checked) {
-                                items.instructionsVisible = false
-                                Activity.selectmode()
-                            }
+                            items.instructionsVisible = instructionsBox.checked
+                            Activity.selectmode()
                         }
-
                     }
 
                     GCDialogCheckBox {
@@ -147,19 +141,20 @@ ActivityBase {
                         text: qsTr("Score visible")
                         checked: items.scoreVisible
                         onCheckedChanged: {
-                            if(scoreBox.checked) {
-                                items.scoreVisible = true
-                                Activity.selectmode()
-                            }
-                            else if(!scoreBox.checked) {
-                                items.scoreVisible = false
-                                Activity.selectmode()
-                            }
-
+                            items.scoreVisible = scoreBox.checked
+                            Activity.selectmode()
                         }
-
                     }
 
+                    GCDialogCheckBox {
+                        id: categoryImageBox
+                        width: instructionsBox.width
+                        text: qsTr("Category image visible")
+                        checked: items.categoryImageChecked
+                        onCheckedChanged: {
+                            items.categoryImageChecked = categoryImageBox.checked
+                        }
+                    }
                 }
             }
             onLoadData: {
--- a/src/activities/categorization/Categorization.qml
+++ b/src/activities/categorization/Categorization.qml
@@ -68,8 +68,7 @@ ActivityBase {
             property bool categoryImageChecked: categoryImageVisible
             property bool scoreChecked: scoreVisible
             property bool instructionsChecked: instructionsVisible
-            property bool iAmReadyChecked: ((!instructionsVisible && !scoreVisible) || (!instructionsVisible && scoreVisible))
-            property bool displayUpdateDialogAtStart: true            
+            property bool displayUpdateDialogAtStart: true
             property var details
             property bool categoriesFallback
             property alias file: file
@@ -100,10 +99,10 @@ ActivityBase {
         MenuScreen {
             id: menuScreen
 
-        File {
-            id: file
-            onError: console.error("File error: " + msg);
-        }
+            File {
+                id: file
+                onError: console.error("File error: " + msg);
+            }
         }
 
         Directory {
@@ -133,8 +132,8 @@ ActivityBase {
                         checked: items.instructionsVisible
                         onCheckedChanged: {
                             items.instructionsVisible = instructionsBox.checked
-                            menuScreen.iAmReady.visible = ((!items.scoreVisible && !items.instructionsVisible) || (!items.instructionsVisible && items.scoreVisible))
                         }
+
                     }
 
                     GCDialogCheckBox {
@@ -144,7 +143,6 @@ ActivityBase {
                         checked: items.scoreVisible
                         onCheckedChanged: {
                             items.scoreVisible = scoreBox.checked
-                            menuScreen.iAmReady.visible = ((!items.scoreVisible && !items.instructionsVisible) || (!items.instructionsVisible && items.scoreVisible))
                         }
                     }
 
@@ -155,14 +153,12 @@ ActivityBase {
                         checked: items.categoryImageChecked
                         onCheckedChanged: {
                             items.categoryImageVisible = categoryImageBox.checked
-                            menuScreen.iAmReady.visible = ((!items.scoreVisible && !items.instructionsVisible) || (!items.instructionsVisible && items.scoreVisible))
-
                         }
                     }
                 }
             }
             onLoadData: {
-                if(dataToSave) {
+                if(dataToSave && dataToSave["scoreVisible"]) {
                     items.scoreVisible = dataToSave["scoreVisible"]
                     items.instructionsVisible = dataToSave["instructionsVisible"]
                     items.categoryImageVisible = dataToSave["categoryImageVisible"]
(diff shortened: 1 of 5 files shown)


divyam3897:
You would probably need one more check box for the category image (one at top right) to handle it. We can probably also handle everything with the combinations of score and instructions and image and probably not even need the easy, medium and expert modes?

amitsagtani97:
- [x] Added Category Image Check box

divyam3897:
I haven't had a look at the code (I just tested it for now). Here are the issues to be fixed:
1. Unchecking the intructions do not hide them.
2. not always but I reproduced it many a times -> disable the instructions and enable the score and images, I didn't see any images
3. The zone color changes still when image completely enters (it could be better if it changes when image just touches the zone, just enters).
4. The options last checked/unchecked is not saved in config. (Chose an option, close categorization and then come back, I couldn't find the one I chose).

divyam3897:
Had the following issues:
1. In medium mode and expert mode (the images overwrite the category image) : https://pasteboard.co/GCKTNWu.png
2. In medium mode, if I exit the config it is in medium mode but if I return from a level it shows me I am ready button.
3. Theconfig saving is better but it still had some issues (not reproducible always).
4.  I got  the following warnings:
```
qrc:/gcompris/src/activities/categorization/Categorization.qml:171: Error: Cannot assign [undefined] to bool
QFont::setPointSizeF: Point size <= 0 (0.000000), must be greater than 0
```
5. I am not sure if there should be so many modes for expert (I see I am ready in various situations, may be in many cases it would not be required, probably we can keep them simple in those cases) It needs to be checked with everyone regarding what would be suitable :)

src/activities/categorization/CategoryReview.qml
```diff
@@ -76,7 +75,7 @@ Item {
             x: leftScreen.width + middleScreen.width + 0.01 * parent.width
             z: 2
             anchors.top: categoryBackground.top
-            anchors.topMargin: items.mode != "expert" ? rootItem.categoryImage.height + 0.027 * rightScreen.height : 0.05 * categoryBackground.height
+            anchors.topMargin: !(!items.instructionsVisible && !items.scoreVisible) ? rootItem.categoryImage.height + 0.027 * rightScreen.height : 0.05 * categoryBackground.height
```
divyam3897: anchors.topMargin: items.instructionVisible && items.scoreVisible?

src/activities/categorization/Categorization.qml
```diff
@@ -62,26 +62,30 @@ ActivityBase {
...
-            property bool categoryImageChecked: (mode === "easy" || mode === "medium")
-            property bool scoreChecked: (mode === "easy")
-            property bool iAmReadyChecked: (mode === "expert")
-            property bool displayUpdateDialogAtStart: true
+            property bool categoryImageVisible: true
+            //property string mode: "easy"
+            property bool categoryImageChecked: categoryImageVisible //(mode === "easy" || mode === "medium")
+            property bool scoreChecked: scoreVisible //(mode === "easy" || mode === "noInstructionsScoreVisible")
+            property bool instructionsChecked: instructionsVisible //(mode === "easy" || mode === "medium")
+           // property bool iAmReadyChecked: ((!instructionsVisible && !scoreVisible) || (!instructionsVisible && scoreVisible))
```
divyam3897: Remove the comments.

src/activities/categorization/Zone.qml
```diff
@@ -40,15 +40,13 @@ Flow {
         model: zoneModel
         Item {
             id: item
-            width: horizontalLayout ? middleZone.width * 0.32 : middleZone.width * 0.49
-            height: horizontalLayout ? categoryBackground.height * 0.2 : categoryBackground.height * 0.15
+            width: middleScreen.width*0.32
```
divyam3897: why did you change the width and height and removed the condition for horizontalLayout?

src/activities/categorization/Zone.qml
```diff
@@ -40,15 +40,13 @@ Flow {
...
+            height: categoryBackground.height * 0.2
             opacity: 1
             Image {
                 id: image
-                fillMode: Image.PreserveAspectFit
-                sourceSize.width: horizontalLayout ? middleZone.width * 0.32 : middleZone.width * 0.49
-                width: sourceSize.width
-                height: sourceSize.width
+                width: middleScreen.width*0.28
+                height: categoryBackground.height * 0.15
```
divyam3897: same as above why did you change the width and height?

src/activities/categorization/CategoryReview.qml
petitlapin: just for info (amit fixed it in another way), the simplification is "items.instructionVisible || items.scoreVisible" with a OR

src/activities/categorization/Categorization.qml
```diff
@@ -62,26 +62,29 @@ ActivityBase {
...
             property alias dialogActivityConfig: dialogActivityConfig
-            property string mode: "easy"
+            property bool scoreVisible: true
             property bool instructionsVisible: true
-            property bool categoryImageChecked: (mode === "easy" || mode === "medium")
-            property bool scoreChecked: (mode === "easy")
-            property bool iAmReadyChecked: (mode === "expert")
-            property bool displayUpdateDialogAtStart: true
+            property bool categoryImageVisible: true
+            property bool categoryImageChecked: categoryImageVisible
```
petitlapin: do you need both?

src/activities/categorization/Categorization.qml
```diff
@@ -62,26 +62,29 @@ ActivityBase {
...
+            property bool scoreChecked: scoreVisible
+            property bool instructionsChecked: instructionsVisible
+            property bool iAmReadyChecked: ((!instructionsVisible && !scoreVisible) || (!instructionsVisible && scoreVisible))
```
petitlapin: !instructionsVisible only?
Why don't you use it for iAmReady visible which is the same if I'm not wrong?

src/activities/categorization/Categorization.qml
```diff
@@ -123,73 +122,73 @@ ActivityBase {
...
-                        exclusiveGroup: configOptions
+                        text: qsTr("Instructions visible")
+                        checked: items.instructionsVisible
                         onCheckedChanged: {
-                            if(easyModeBox.checked) {
-                                items.mode = "easy"
-                                menuScreen.iAmReady.visible = false
-                            }
+                            items.instructionsVisible = instructionsBox.checked
+                            menuScreen.iAmReady.visible = ((!items.scoreVisible && !items.instructionsVisible) || (!items.instructionsVisible && items.scoreVisible))
```
petitlapin: it should be done automatically with the binding, no need to reset it (and break the binding)

src/activities/categorization/Categorization.qml
```diff
@@ -123,73 +122,73 @@ ActivityBase {
...
+                        width: instructionsBox.width
+                        text: qsTr("Score visible")
+                        checked: items.scoreVisible
                         onCheckedChanged: {
-                            if(mediumModeBox.checked) {
-                                items.mode = "medium"
-                                menuScreen.iAmReady.visible = false
-                            }
+                            items.scoreVisible = scoreBox.checked
+                            menuScreen.iAmReady.visible = ((!items.scoreVisible && !items.instructionsVisible) || (!items.instructionsVisible && items.scoreVisible))
```
petitlapin: no need

src/activities/categorization/Categorization.qml
```diff
@@ -123,73 +122,73 @@ ActivityBase {
...
+
                         }
                     }
                 }
             }
             onLoadData: {
-                if(dataToSave && dataToSave["mode"])
-                    items.mode = dataToSave["mode"]
+                if(dataToSave) {
```
petitlapin: you need to at least add a: && dataToSave["scoreVisible"]
else it will fail at first access when you don't have it already stored

src/activities/categorization/Categorization.qml
```diff
@@ -123,73 +122,73 @@ ActivityBase {
...
             onSaveData: {
                 dataToSave["data"] = Activity.categoriesToSavedProperties(dataToSave)
-                dataToSave["mode"] = items.mode
                 dataToSave["displayUpdateDialogAtStart"] = items.displayUpdateDialogAtStart ? "true" : "false"
+                dataToSave["scoreVisible"] = items.scoreVisible
+                dataToSave["instructionsVisible"] = items.instructionsVisible
+                dataToSave["categoryImageVisible"] = items.categoryImageVisible
+            }
+            onClose: {
+                menuScreen.iAmReady.visible = ((!items.scoreVisible && !items.instructionsVisible) || (!items.instructionsVisible && items.scoreVisible))
```
petitlapin: should not be needed

src/activities/categorization/CategoryReview.qml
```diff
@@ -218,7 +215,7 @@ Item {
     }
 
     function stop() {
-        if(items.mode == "expert")
+        if((!items.instructionsVisible && !items.scoreVisible) || (!items.instructionsVisible && items.scoreVisible))
```
petitlapin: !items.instructionsVisible only?

src/activities/categorization/categorization.js
```diff
@@ -202,16 +204,25 @@ function getCategoryLevels() {
...
+     if((items.scoreVisible && items.instructionsVisible) || (items.instructionsVisible && !items.scoreVisible)) {
         items.details = lessons[index].map(function(ele) {
             return { "instructions": ele.instructions, "image": ele.image,
                 "numberOfGood": ele.maxNumberOfGood, "numberofBad": ele.maxNumberOfBad,
                 "categoryImages": ele.levelImages ,"good": ele.good,
                 "bad": ele.bad ,"prefix": ele.prefix }
         });
     }
+
+     else if(!items.instructionsVisible && items.scoreVisible ) {
```
petitlapin: remove space before )

petitlapin:
as discussed on irc

PR #207, opened 2017-12-27
Categorization flowers

gupta2140:
Addition of images, README and dataset is completed.


src/activities/categorization/resource/board/category_flowers.qml
```diff
@@ -0,0 +1,116 @@
+/* GCompris
+ *
+ * Copyright (C) 2017 Aman Kumar Gupta <gupta2140@gmail.com>
```
divyam3897: 2017 -> 2018

gupta2140: I have to update this PR after image dataset is merged into gcompris-words repo. So, i'll make all the changes at once :)

src/activities/categorization/resource/board/category_flowers.qml
```diff
@@ -0,0 +1,116 @@
...
+import QtQuick 2.6
+
+QtObject {
+    property bool isEmbedded: false
+    property bool allowExpertMode: true
+    property string imagesPrefix: "qrc:/gcompris/data/words/flowers/"
+    property var levels: [
+        {
+            "name": qsTr("Flowers"),
+            "image": imagesPrefix + "rose.jpg",
```
divyam3897: can we have a flowers image instead to represent the category instead of just a rose?

gupta2140: yes, there's a "bunch of flowers" image in the image dataset...we can use that one.

gupta2140:
ahh, ok i'll change them when changing the datasets for new path of images :)

petitlapin:
Merged in https://commits.kde.org/gcompris/4bf1afec5bb330c44d3ede36f00f413444564830

Thank you,

Johnny

PR #208, opened 2017-12-29
core, add confirmation dialog on exit

amitsagtani97:
Patch for [Task5013](https://phabricator.kde.org/T5013)
 
--> Shows confirmation dialog on exiting an activity or closing an activity.

### DIFF
--- a/src/activities/menu/Menu.qml
+++ b/src/activities/menu/Menu.qml
@@ -47,6 +47,9 @@ ActivityBase {
     id: activity
     focus: true
     activityInfo: ActivityInfoTree.rootMenu
+    // set to true when returned to home from a dialog.
+    property bool returnFromDialog: false
+
     onBack: {
         pageView.pop(to);
         // Restore focus that has been taken by the loaded activity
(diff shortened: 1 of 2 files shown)


src/activities/menu/Menu.qml
```diff
@@ -47,6 +47,9 @@ ActivityBase {
     id: activity
     focus: true
     activityInfo: ActivityInfoTree.rootMenu
+    // set to true when returned to home from a dialog.
+    property bool returnFromDialog: false
```
amitsagtani97: This variable is needed to handle the cases when we close a dialog.
home() is called when we close a dialog, so to prevent the confirmation dialog to pop up, in that case, I have used this variable to handle this.

src/activities/menu/Menu.qml
```diff
@@ -58,17 +61,33 @@ ActivityBase {
...
             pageView.pop();
             // Restore focus that has been taken by the loaded activity
             if(pageView.currentItem == activity)
                 focus = true;
         }
+        // A confirmation dialog to quit the current activity is popped.
+        else {
+            Core.showMessageDialog(parent,
+                                   qsTr("Do you really want to quit?"),
+                                   qsTr("Yes"), function() { pageView.pop();
```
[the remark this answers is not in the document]
petitlapin: this code is copied/pasted 3 times, can be in a function instead

src/activities/menu/Menu.qml
```diff
@@ -47,6 +47,9 @@ ActivityBase {
     id: activity
     focus: true
     activityInfo: ActivityInfoTree.rootMenu
+    // set to true when returned to home from a dialog.
+    property bool returnFromDialog: false
+
     onBack: {
```
petitlapin: I think it should also be added there (I don't have mouse with back button but I think it will trigger this signal)

amitsagtani97: This signal is triggered only on back pressed from phones or other touch devices. Currently, back button from the mouse is not enabled.

src/activities/menu/ConfigurationItem.qml
```diff
@@ -213,6 +213,15 @@ Item {
             }
         }
 
+        GCDialogCheckBox {
+            id: enableExitDialogBox
+            text: qsTr("Enable activity Exit confirmation dialog")
```
petitlapin: Confirm before leaving current activity

PR #210, opened 2018-01-18
Control the hose pipe: Fix layout for vertical mode [T7610]

parth-p:


petitlapin:
Hi,

thank you for your PR.

There are multiple things to change in this activity regarding this issue:
* the fireman and the fire are hidden by the bar: https://pasteboard.co/H3Z5mG6.png (can probably be fixed using anchors)
* when resizing only the width, the pipe is not refreshed (so it goes outside the screen).

Thank you

parth-p:
@petitlapin , for the hidden fire and fire man do i need to change their y coordinates in order to see them properly ?

petitlapin:
No, you can use anchors property directly instead of setting x/y (I let you dig :)).
For the ratio of the screen, I just switched from fullscreen to windowed mode (ctrl+F) and resized it

parth-p:
@petitlapin hi , i made some changes using anchors. Suggest if any more change is required.

petitlapin:
Commited in https://commits.kde.org/gcompris/2b885eb95cf449156bc10597cec6b259ce35d338 with some changes, please verify them :).

Thank you,

Johnny

PR #215, opened 2018-01-28
letter_in_word: Fixing overriding issues with images when expanded [T5935]

parth-p:
For the overriding issue I increased the distance b/w two rows and reduced the size of the scaled object.


petitlapin:
As said on the task (https://phabricator.kde.org/T5935): https://pasteboard.co/H64MuWL.png
This activity needs more work to better place the elements (like for colors activity).

parth-p:
As I suggested for a tick instead of scaling on click, I made some changes . Here is a screenshot of how it looks with this solution : https://pasteboard.co/H7lUxpn.png 
the image of tick and position can be changes if this looks good . Please suggest...

PR #220, opened 2018-02-22
core, Fix hidden text in help [T8049]

darkHarry:
This PR fixes the partly hidden text of Help section in vertical mode of a few activities where the title is in more than one line.
Changed the height of the Rectangle to consider the change in height of title.
Files changed : DialogBackground.qml


petitlapin:
Committed in https://commits.kde.org/gcompris/8f65adc2257549009f867ff7eef092c661fac762
Thank you,
Johnny

PR #224, opened 2018-03-05
tangram, fix few out placed models, remove multiple solution levels

darkHarry:
This Patch removes levels with multiple solutions and fixes the positioning of a few out placed models.


src/activities/tangram/dataset.js
```diff
@@ -92,7 +92,7 @@ var dataset = [
                     },
                 ]
             },
-            {
+                {
```
petitlapin: change not needed?

src/activities/tangram/dataset.js
```diff
@@ -942,6 +942,7 @@ var dataset = [
                     },
                 ]
             },
+            /*
```
petitlapin: you can directly remove them instead of commenting them

src/activities/tangram/Tangram.qml
```diff
@@ -132,8 +132,8 @@ ActivityBase {
                     sourceSize.width: modelData.width * background.playWidth
                     sourceSize.height: modelData.height * background.playWidth
                     z: index
-                    rotation: modelData.flipping ? 360 - modelData.rotation : modelData.rotation
-                    mirror: modelData.flipping ? true : false
+                    rotation: modelData.rotation
```
petitlapin: do you have an example of level where it applies? I'm not sure to understand the change

src/activities/tangram/Tangram.qml
```diff
@@ -132,8 +132,8 @@ ActivityBase {
...
+                    mirror: modelData.flipping ? true : false
```
petitlapin: can you replace the character with a normal space? I'm not sure on how it went there on the first place but let's use normal spaces.

src/activities/tangram/Tangram.qml
darkHarry: 11 levels created this problem with parallelogram. *screenshot*
The rotation of modelTan was out of place and the solution was being accepted.

src/activities/tangram/Tangram.qml
darkHarry: it comes whenever i change the line above it in qt creator. Used other editor and didn't happen. :/

Animtim:
Hi, looks like you didn't do exacly what was asked... except for level 11 where it doesn't look the same after your qml change, I don't understand why you had to change the dataset like that?

petitlapin:
committed in https://commits.kde.org/gcompris/0bdda9aa7da4198726ca77f3339bc661873b1e5f

Thank you,

Johnny

PR #226, opened 2018-03-10
Background music: lowered background music volume.

gupta2140:
Lowered background music to be able to hear activity intro voices distinctly and clearly.

### DIFF
--- a/src/core/GCAudio.qml
+++ b/src/core/GCAudio.qml
@@ -227,6 +227,7 @@ Item {
 
     Audio {
         id: audio
+        muted: gcaudio.muted
         onError: {
             // This file cannot be played, remove it from the source asap
             source = ""
--- a/src/core/GCAudio.qml
+++ b/src/core/GCAudio.qml
@@ -100,14 +100,6 @@ Item {
      */
     signal done
 
-    property var muteChangeHandler: function () { muted ? volume = 0 : volume = 1 }
-
-    /**
-     *  When mute is changed we set the volume to 0 to mute a potential playing
-     * sound.
-     */
-    onMutedChanged: muteChangeHandler()
-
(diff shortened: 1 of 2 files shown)


src/core/main.qml
```diff
@@ -129,13 +129,14 @@ Window {
     GCAudio {
         id: backgroundMusic
         muted: !ApplicationSettings.isBackgroundMusicEnabled
+        volume: 0.53
```
petitlapin: why not 0.5 directly?
can you make a readonly default value and assign it to the value instead of putting it here and below?

Then, it should be good to go, you can directly commit on the branch.
Thank you

Johnny

ps: as a more long time contributor (and you already have an account) please prefer differential for patches :)

gupta2140: sure, I'll prefer the differential for patches from next tasks.

src/core/GCAudio.qml
```diff
@@ -100,14 +100,6 @@ Item {
      */
     signal done
 
-    property var muteChangeHandler: function () { muted ? volume = 0 : volume = 1 }
-
-    /**
-     *  When mute is changed we set the volume to 0 to mute a potential playing
-     * sound.
-     */
-    onMutedChanged: muteChangeHandler()
```
gupta2140: I don't think there's any more need for these two statements, as now onMutedChanged signal can directly be used externally wherever required.

petitlapin:
ok, good to go :).

Thanks,

Johnny

gupta2140:
Thank you! :)

Commit link: https://cgit.kde.org/gcompris.git/commit/?h=backgroundMusic&id=fecd920d041718112ebef00167de0f9bc317bdd2

PR #227, opened 2018-03-23
PR script, add script and tests

darkHarry:
This Pull Request contains first version of Pull Request Checker Script.

Checks and warns if qsTr is not used in a property (eg: text: "hello world").
Warns for not updated Copyrights
Checks if ActivityInfo.qml file has "createdInVersion" present

Test against a file having all possible errors.


src/Gemfile
```diff
@@ -0,0 +1,3 @@
+source 'https://rubygems.org'
```
petitlapin: is it possible to have a version without having to install external modules, only ruby?
We can live without the commit name check for now and the colors

src/checkActivities.rb
```diff
@@ -0,0 +1,143 @@
+require 'colorize'
+require 'git'
+
+module ActivityCheck
+
+    def self.initialize(activityDirName)
+        @activityDirName = activityDirName
+
+        # create js file name array
+        @jsFilesArray = Dir.glob("../src/activities/" + @activityDirName + "/*.js")
```
petitlapin: it should be recursive inside the folder (for example, we can have for some activities, @activityDirName + "/resource/boards/*qml" files

src/checkActivities.rb
```diff
@@ -0,0 +1,143 @@
...
+            puts "\nFilePath: " + jsFile
+
+            File.foreach(jsFile) do |line|
+
+                if line.include?("THE GTK VERSION AUTHOR")
+                    print_error "Replace original GTK VERSION AUTHOR label by your own name"
+                end
+
+                if line.include?("THE GTK VERSION AUTHOR")
+                    print_warning "Replace original QT Quick copyright label email by your own email address"
```
petitlapin: Qt not QT :)

src/checkActivities.rb
```diff
@@ -0,0 +1,143 @@
...
+                end
+            end
+
+            File.foreach(qmlFile) do |line|
+
+                if File.basename(qmlFile) == "ActivityInfo.qml"
+                    activityInfo_QmlFile line, @activityDirName
+                end
+
+                if line.include?("THE GTK VERSION AUTHOR")
```
petitlapin: it's the same code as for js, can it be factorised?

src/checkactivitytest.rb
```diff
@@ -0,0 +1,47 @@
+require_relative "checkActivities"
+require "test/unit"
+
+class TestCheckActivity < Test::Unit::TestCase
+
+    @@outputFile = "testoutput.txt"
+    @@expecOutputFile = "regressiontestoutput.txt"
+
+    def change_stdout(file)
+        if $stdout == STDOUT then $stdout = file
```
petitlapin: I'm not sure to understand why you just don't set $stdout to file. And if you want to revert to the real stdout, call the function with STDOUT?

src/Gemfile
darkHarry: but i thought commit name check was required. Will it be required later on ?

petitlapin: as we are trying to push contributors to use phabricator instead of github, names should be fine later.
It is not one of the most important change and we can do without it.

src/checkActivities.rb
darkHarry: yes.

src/Gemfile
petitlapin: file no more needed?

src/Gemfile.lock
```diff
@@ -0,0 +1,15 @@
+GEM
```
petitlapin: file no more needed?

src/regressiontestoutput.txt
```diff
@@ -0,0 +1,23 @@
...
+FilePath: ../src/activities/testactivity/Testactivity.qml
+	 ERROR       Replace original GTK VERSION AUTHOR label by your own name
+	 WARNING     Replace original QT Quick copyright label email by your own email address
+	 WARNING     "testactivity activity" qsTr may not be used
+
+FilePath: ../src/activities/testactivity/ActivityInfo.qml
+	 WARNING     difficulty: 1 may not be updated.
+	 ERROR       author: "Your Name &lt;yy@zz.org&gt;" may not be updated.
+	 WARNING     "Your Name &lt;yy@zz.org&gt;" qsTr may not be used
+	 ERROR       demo: true may not be updated.
```
petitlapin: this one should always be true actually (false is for activities on the paying version, and we won't add anymore paying activities).
It would be better to check the opposite.

src/Gemfile
petitlapin: file to be removed?

src/Gemfile.lock
petitlapin: file to be removed?

src/regressiontestoutput.txt
```diff
@@ -0,0 +1,23 @@
...
+
+FilePath: ../src/activities/testactivity/Testactivity.qml
+	 ERROR       Replace original GTK VERSION AUTHOR label by your own name
+	 WARNING     Replace original QT Quick copyright label email by your own email address
+	 WARNING     "testactivity activity" qsTr may not be used
+
+FilePath: ../src/activities/testactivity/ActivityInfo.qml
+	 WARNING     difficulty: 1 may not be updated.
+	 ERROR       author: "Your Name &lt;yy@zz.org&gt;" may not be updated.
+	 WARNING     "Your Name &lt;yy@zz.org&gt;" qsTr may not be used
```
petitlapin: there are a lot of false positive for this warning.
We can try to simplify it later or at least add a list of false positive to skip

src/Gemfile
darkHarry: No, this file is used to install Color and Git modules and since they aren't required for the moment i've removed the Gemfiles.

src/Gemfile.lock
darkHarry: Not needed.

src/checkactivitytest.rb
darkHarry: I'm sorry, i actually don't remember why i did that. :P 
I've changed it now.

src/regressiontestoutput.txt
darkHarry: okay, didn't know that.

src/regressiontestoutput.txt
darkHarry: yes, this is what i was concerned about too. Some other way has to be found for checking qsTr to avoid these false positives.

src/checkActivities.rb
```diff
@@ -0,0 +1,115 @@
+module ActivityCheck
```
petitlapin: missing header with license

src/checkActivities.rb
```diff
@@ -0,0 +1,136 @@
...
+    def self.qsTr_use(lineStr, lineNumber)
+        matchStr = /([^ ]+): "[^"]+"/.match(lineStr)
+        if matchStr
+            print_warning "line:#{lineNumber} #{matchStr} qsTr may not be used"
+        end
+    end
+
+    # checks copyrights updated
+    def self.activityInfo_QmlFile(lineStr, directoryName)
+        valuesToTestArray = ["author: \"Your Name &lt;yy@zz.org&gt;\"", "name: \"test/Test.qml\"", "difficulty: 1", "icon: \"test/test.svg\"",
```
petitlapin: for "demo: false", can you only display the warning if Bruno Coudoin is not the author?

src/checkActivities.rb
```diff
@@ -0,0 +1,136 @@
...
+
+    # to print warning
+    def self.print_warning(wrn)
+        puts "\t WARNING     #{wrn}"
+    end
+
+    # checks import version
+    def self.check_import_version(lineStr)
+        # add modules and versions in hash
+        moduleVersionHash = { "QtQuick" => "2.6", "GCompris" => "1.0" }
```
petitlapin: moduleVersionHash = { "QtQuick" => "2.6", "QtGraphicalEffects" => "1.0",
                            "QtQuick.Controls" => "1.5", "QtQml" => "2.2",
                            "QtQuick.Controls.Styles" => "1.4",
                            "QtQuick.Particles" => "2.0", "QtSensors" => "5.0",
                            "QtMultimedia" => "5.0",
                            "GCompris" => "1.0", "Box2D" => "2.0" }

src/checkActivities.rb
darkHarry: i don't understand. Why is that so ? every activity he developed does not have a demo ?

src/checkActivities.rb
darkHarry: Thanks : )

src/checkActivities.rb
```diff
@@ -0,0 +1,136 @@
+module ActivityCheck
+
```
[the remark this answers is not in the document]
darkHarry: sure.

src/checkActivities.rb
petitlapin: Only the activities developed by Bruno should have demo: false

petitlapin:
work merged in checkActivitiesScripts branch and to be continued before merge

PR #240, opened 2018-11-11
core, add configuration modes for Domino component

AkshayCHD:
Two modes have been added(for simple and roman numbers) and a configuration menu in reversecount activity has been created.
image
image

image

### DIFF
--- a/src/activities/reversecount/Reversecount.qml
+++ b/src/activities/reversecount/Reversecount.qml
@@ -42,6 +42,7 @@ ActivityBase {
         signal stop
 
         Component.onCompleted: {
+            dialogActivityConfig.getInitialConfiguration()
             activity.start.connect(start)
             activity.stop.connect(stop)
         }
@@ -59,6 +60,7 @@ ActivityBase {
             property alias tux: tux
             property alias fishToReach: fishToReach
             property int clockPosition: 4
+            property string mode: "dot"
         }
 
         onStart: { Activity.start(items) }
@@ -162,13 +164,18 @@ ActivityBase {
 
         Bar {
             id: bar
-            content: BarEnumContent { value: help | home | level }
+            content: BarEnumContent { value: help | home | level | config}
             onHelpClicked: {
                 displayDialog(dialogHelp)
             }
             onPreviousLevelClicked: Activity.previousLevel()
             onNextLevelClicked: Activity.nextLevel()
             onHomeClicked: activity.home()
+            onConfigClicked: {
+                dialogActivityConfig.active = true
+                dialogActivityConfig.setDefaultValues()
+                displayDialog(dialogActivityConfig)
+            }
         }
 
         Image {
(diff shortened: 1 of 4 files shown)


src/activities/reversecount/Reversecount.qml
```diff
@@ -59,6 +59,7 @@ ActivityBase {
             property alias tux: tux
             property alias fishToReach: fishToReach
             property int clockPosition: 4
+            property int mode: 1
```
petitlapin: Harcoded numbers shouldn't be used for mode: it's difficult to track to what they really correspond.
In the current case, it would be quite easy to have a difference between the configuration value and the domino value if we change one or the other and it would be complicated to spot

src/core/Domino.qml
```diff
@@ -59,6 +59,12 @@ Flipable {
     property color backColor: "white"
     property color pointColor: "black"
 
+    // Define the mode/type of domnino
```
petitlapin: domino

src/core/Domino.qml
```diff
@@ -59,6 +59,12 @@ Flipable {
...
+    // 1 is for Dot representation
+    // 2 is for Number  representation
```
petitlapin: double space

src/core/Domino.qml
```diff
@@ -76,16 +82,53 @@ Flipable {
...
             height: parent.height
             color: flipable.pointColor
             borderColor: "black"
             borderWidth: 0
             radius: parent.height * 0.25
             valueMax: flipable.valueMax
-            onValueChanged: if(flipEnabled) flipable.flipped = !flipable.flipped
+            onValueChanged: {
+                if(flipEnabled) flipable.flipped = !flipable.flipped
+                number1.value = this.value;
```
petitlapin: no need of "this"

src/activities/reversecount/Reversecount.qml
```diff
@@ -213,8 +219,57 @@ ActivityBase {
...
+                    ]
+                    Flow {
+                        id: flow
+                        spacing: 5
+                        width: dialogActivityConfig.width
+                        GCComboBox {
+                            id: modeBox
+                            model: availableModes
+                            background: dialogActivityConfig
+                            label: qsTr("Select Domino Type")
```
petitlapin: would "representation" be better than "type"?

src/core/Domino.qml
```diff
@@ -76,16 +82,53 @@ Flipable {
         border.width: flipable.borderWidth
         radius: flipable.radius
 
-        DominoNumber {
-            id: number1
+        DominoDot {
```
petitlapin: I don't like the idea of triplicate the dominoes in the code + having 3 different dominoes with a lot of copy/paste.
Is there a way to simplify this?
Here there are dependencies between the dominoes (onValueChanged), and it seems except "visible" property, almost all the other property are copy/pasted for 6 items

src/core/DominoRoman.qml
```diff
@@ -0,0 +1,162 @@
...
+    property bool isClickable: true // Default value
+
+    /**
+     * type:GCAudio
+     * To play sound and audio effects.
+     */
+    property GCSfx audioEffects
+
+    function getText() {
+        if(item.value == 1)
```
petitlapin: it can probably be handled this via an array to gain some readability

src/core/DominoNumber.qml
```diff
@@ -78,6 +84,26 @@ Item {
...
+            return item.value
+        var romans = ["", "I", "II", "III", "IV", "V", "VI", "VII", "VIII", "IX"]
+        return romans[item.value]
+    }
+
+    GCText {
+        id: numberText
+        y: 0
+        visible: (item.mode == "number" || item.mode == "roman")
+        fontSize: ((item.mode == "number") ? 30 : (item.value == 8) ? 20 : (item.value == 7 || item.value == 3) ? 25 : 30)
```
petitlapin: you can probably use "fontSizeMode: Text.Fit" instead of handling it manually?

src/core/DominoNumber.qml
```diff
@@ -78,6 +84,26 @@ Item {
...
+    function getText() {
+        if(mode == "number")
+            return item.value
+        var romans = ["", "I", "II", "III", "IV", "V", "VI", "VII", "VIII", "IX"]
+        return romans[item.value]
+    }
+
+    GCText {
+        id: numberText
+        y: 0
```
petitlapin: no need of setting it if 0, it is the default value

src/core/DominoNumber.qml
```diff
@@ -78,6 +84,25 @@ Item {
      */
     property GCSfx audioEffects
 
+    function getText() {
```
petitlapin: you can directly have a "readonly property var romans: ["", "I", "II", "III", "IV", "V", "VI", "VII", "VIII", "IX"] and have below text:
text: mode == "number" ? item.value : romans[item.value]

I'll change this directly during the merge, no need to update this PR

petitlapin:
Hi, it's commited under https://commits.kde.org/gcompris/0462d4cc5e4f9f0a11b4cd51162f51120a165946
Thank you for the PR

PR #249, opened 2019-12-08
Updated planet information

restouffer:
My daughter was very troubled to find that the test claimed Venus had two moons, so we looked over the data together to make sure everything looked correct.

### DIFF
--- a/src/activities/solar_system/Dataset.js
+++ b/src/activities/solar_system/Dataset.js
@@ -56,8 +56,8 @@ function get() {
                     "realImg": "qrc:/gcompris/src/activities/solar_system/resource/mercury.png",
                     "bodyName": qsTr("Mercury"),
                     "bodySize": 0.12,
-                    "temperatureHint": qsTr("The maximum temperature on Earth is 58 °C."),
-                    "lengthOfYearHint": qsTr("The length of a year on Venus is 225 days."),
+                    "temperatureHint": qsTr("The maximum temperature on Mercury is 427 °C."),
+                    "lengthOfYearHint": qsTr("The length of a year on Mercury is 88 days."),
                     "levels": [
                         { 	// sub-level 1
                             "question": qsTr("At which position is Mercury in the Solar System?"),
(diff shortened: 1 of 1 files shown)


petitlapin:
HI,

thank you for the corrections!
The aim for the hints were to give information about other planets that could help finding the answers for the actual planet.
If it is misleading (we are always looking for children feedbacks to help us improve the software), we can change it as you suggest.

For accepting the pull request, I would need to change your commit name to something in 2 parts: "firstName lastName <email>" as the main repo is in KDE and it's a requirement to have the author name this way. Can you give me either your full name or a 2-parts pseudonyme if you don't want to give your full name?

Johnny

restouffer:
The hints were confusing. I didn't see them appear in the solar system activity, so I wasn't sure how they were used. I can revert those if you want. Maybe at some point in the future those can be moved to their own section. It might even be possible to break out the facts and generate the hints and quiz questions algorithmically.

My name is Bob Stouffer. You can use that in the commit message.

petitlapin:
Thank you,
I'm asking some teachers that contribute to the software for their opinion about the hints.
You can see them when you click on the "Hint" button when you are replying to the question for a planet.

For example, if you are replying about the temperature on Venus, clicking on the Hint will tell you that the maximum temperature on Earth is 58 °C, and the closer the planet to the Sun is, the hotter it is. And you can see there the solar system map where Venus is closer than Earth so it indicates it is hotter.

I'll handle the pull request depending on their answer, no need to update it.

Johnny

petitlapin:
Here is some feedback:
"The hints are only useful for 2 series of questions (temperature and revolution time). We don't have for helping knowing how much time it takes to do a revolution. It would be better they are only displayed for those questions."

I opened a task in our tracker: https://phabricator.kde.org/T12339.

So for now, I'll just update the answers to have the good ones, the hint system will be reworked when with the task.

I've commited the changes in https://commits.kde.org/gcompris/2531761f6ae2db9a5b09fd17a0413c2af54ae9a1
Thank you again!

PR #252, opened 2020-03-10
new activity added

anuj952:
This is my new activity as per the guidlines for exercise for new contributors . The activity has name anuj_activity


src/activities/anuj_activity/Anuj_activity.qml
```diff
@@ -0,0 +1,202 @@
...
+                }
+                height : parent.height * 0.8
+              }
+
+
+
+            Image {
+                id: rocket
+                x : 0
+                y : 300
```
petitlapin: positions should not be hardcoded, the resolution of the device may change (depending on a phone, table, screen...)

src/activities/anuj_activity/Anuj_activity.qml
```diff
@@ -0,0 +1,202 @@
...
+
+            Rectangle {
+                id : sun
+                width : 200
+                height : 200
+                x : 1500
+                y : 175
+
+                Image {
+                    id : sunimage
```
petitlapin: there should not be a space before :

anuj952:
Thanks for the review , I will use proper conventions in future

PR #275, opened 2021-12-15
Ubuntu Touch: Updates

lduboeuf:


src/core/GCAudio.qml
```diff
@@ -214,7 +214,10 @@ Item {
             source = ""
             gcaudio.done()
         } else {
-            source = ""
+            // on Ubuntu Touch, emptying the source result in an Audio error which triggers that method again endlessly
```
petitlapin, line 217: should it be the same on line 140?

lduboeuf: idk, but haven't spot any audio issue with it. Need to dig a bit to know the difference between those two

petitlapin:
Hi,

I've pushed both commit in https://invent.kde.org/education/gcompris/-/commit/d8448c704348555d804b236d0c9cdcec0f470a57 and https://invent.kde.org/education/gcompris/-/commit/e06a55df3e732855e23a888f21d1130e7fa1e99f

src/core/GCAudio.qml
petitlapin: If I remember correctly, "play" is called to play the sound directly, while the other one is when you append audio to play after the current one has finished.

Intro audio and when you lose on an activity (for example reversecount) will play using "play". You can lose twice and it should be the same audio (try_again) playing twice in a row

PR #277, opened 2022-03-15
Shutdown rocket anim after crash, allow explosion to roll

kentavv:
Fixes: When rocket crashes the rocket burn would still appear after level reset (until OK was pressed.)
Adds: An explosion rolls laterally the direction the rocket impacted.

### DIFF
--- a/src/activities/land_safe/LandSafe.qml
+++ b/src/activities/land_safe/LandSafe.qml
@@ -90,6 +90,9 @@ ActivityBase {
             property alias bar: bar
             property alias bonus: bonus
             property alias rocket: rocket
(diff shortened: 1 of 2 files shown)


kentavv:
I didn't know what's better for young kids. It's slightly more physically correct and slightly more exciting, but the non-rolling explosion feels more ol' school arcade like. Best wishes

Animtim:
Looks ok for me, thanks!

petitlapin:
Thank you! It has been committed in https://invent.kde.org/education/gcompris/-/commit/398d933feabf42260fb3116bc873ed1416198ae5

PR #278, opened 2022-03-19
Updated with needs of Debian 11 and Raspberry Pi

kentavv:
Instructions for building on fresh Debian 11.2 and Raspberry Pi's. I was not able to figure out QtCreator so added example of cmake alternative.

### DIFF
--- a/HACKING
+++ b/HACKING
@@ -52,17 +52,23 @@ be installed
 
 Installing dependencies on a Debian based systems can be done like this:
 
-    sudo aptitude install libqt5svg5-dev \
-                          qml-module-qtmultimedia qt5-default \
+    sudo apt install libqt5svg5-dev \
+                          qml-module-qtmultimedia \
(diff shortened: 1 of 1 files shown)


petitlapin:
In all cases, cmake is mandatory (QtCreator uses it to build). I've commited https://invent.kde.org/education/gcompris/commit/2b153acb5cdde64b24888f2997c2a7bf7672fb2d for this change.

Thank you!

Johnny

kentavv:
cmake should be tacked onto the apt install line then. It was not installed on fresh Debian 11.2 or Raspberry Pi installs. Best wishes

petitlapin:
I've added it in the line [above](https://invent.kde.org/education/gcompris/-/commit/2b153acb5cdde64b24888f2997c2a7bf7672fb2d#fa65f6dc4a7e1f9f6a001696c59e7fade1feb7ad_53_55) the Qt install to split them. If you think it is better to have them in the same list, I will update it.

PR #282, opened 2022-08-28
Add Spain regions to the geo-country activity

jespino:
This add the country regions for spain (following the instructions provided by the geo-country/MapResources/sources file)


petitlapin:
Thank you for your PR!
Looks good to me!

Our main repository is on invent.kde.org, github is just a mirror. I'll cherry pick the commits there directly.

src/activities/geo-country/resource/board/board18_0.qml
```diff
@@ -0,0 +1,154 @@
...
+   property string instruction: qsTr("Regions of Spain")
+   property var levels: [
+      {
+         "pixmapfile" : "spain/spain_spain.svgz",
+         "type" : "SHAPE_BACKGROUND_IMAGE"
+      },
+      {
+          "pixmapfile": "spain/spain_andalucia.svgz",
+           //: Region of Spain: Andalucía
+           "toolTipText" : qsTr("Andalucía"),
```
petitlapin, line 23: The names should be in English in all the tooltips (and then each translator translates it for its language).

Here, it should be `Andalusia` and same for the other ones. Do you want me to update them when I merge?

jespino: However you prefer, I'm ok doing the changes myself, but whatever you think is easier for you works for me.

petitlapin: If you can do it, it would be better as it would make two pairs of eyes please

jespino: I'll do it tomorrow (i hope) unless you do it yourself before that.

jespino: I found some time, now is everything in english and also I change the title to Autonomous communities that is exactly how are they name in Spain.

petitlapin:
Thank you!
I've pushed it in https://invent.kde.org/education/gcompris/-/commit/baecf57f8cb466333be125cb2cbbb73d247c8b27

PR #291, opened 2023-12-11
Update HACKING to change git to https and add webp dependency

AlexTheAussieNerd:
When installing from these directions i needed to make some changes. 

I needed to use the command "git clone https://github.com/gcompris/GCompris-qt.git" instead of ".....git://...." otherwise the attempt would time out. 

After building i noticed images in the Categorisation activity would not show thumbnail images for "Colours", "Animals" , "Food" etc.  and clicking into Numbers would show the digits and letters (stored as .svg files) correctly, but other files (stored as .webp) were not shown.  running the command "sudo apt install qt5-image-formats-plugins" and recompiling showed the previously invisible images.

### DIFF
--- a/HACKING
+++ b/HACKING
@@ -15,7 +15,7 @@ Source code from GitHub
 -----------------------
 Checkout the source code by cloning it from GitHub
 
-    git clone git://github.com/gcompris/GCompris-qt.git
+    git clone https://github.com/gcompris/GCompris-qt.git
 
 If you want to have a personal fork, in order to commit occasional patches,
 fork it on GitHub then clone your fork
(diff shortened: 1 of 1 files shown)


AlexTheAussieNerd:
Ive also just used the command line to clone my personal fork and couldnt reproduce with the text provided. 

I did get a download started with 'git clone https://github.com/AlexTheAussieNerd/GCompris-qt.git" and request changing of the test in the existing HACKING be changed to "git clone https://github.com/<username>/GCompris-qt.git"

petitlapin:
We have rules in KDE that the author should be in the form "FirstName LastName <mail>". In the commit, it is not the case, making the push failed.

Either I push with my name or if you want the credit (which is understandable as you did the job :)), can you please give me a compliant name for it? Note that the name will be written in GCompris credits directly in the application.

petitlapin:
I've updated the list in https://invent.kde.org/education/gcompris/-/commit/6f6762ff10263a10532e9b1d5c298167d6fba962 with the list we have in the official wiki: https://gcompris.net/wiki/Qt_Quick_development_process#Compilation_prerequisites

Thanks again for pointing them out! It's ok to propose changes here for me and I'll commit in my name if you don't want to use yours, no problem either on my side.

PR #292, opened 2023-12-12
Update HACKING to include openssl dependency and correct command for …

AlexTheAussieNerd:
…forked repositories

I've re-run the install on a different machine and noted some more changes were needed. 

No download would happen on the existing git clone command for forked repositories.

Building as per instructions failed on cmake? saying a dependeny wasn't met for openssl.

### DIFF
--- a/HACKING
+++ b/HACKING
@@ -20,7 +20,7 @@ Checkout the source code by cloning it from GitHub
 If you want to have a personal fork, in order to commit occasional patches,
 fork it on GitHub then clone your fork
 
-    git clone git@github.com:<username>/GCompris-qt.git
+    git clone https://github.com/<username>/GCompris-qt.git
 
 When asking for a pull request, create a new topic branch and commit your
 changes there then open a pull request to gcompris-qt repository on GitHub.
@@ -61,7 +61,7 @@ By default, translations are also built so gettext package must be installed to 
     
 Installing dependencies on a Debian based systems can be done like this:
 
-    sudo apt install libqt5svg5-dev \
+    sudo apt install libqt5svg5-dev libssl-dev\
                           qml-module-qtmultimedia \
                           qml-module-qtgraphicaleffects qt5-qmake qtcreator \
                           qtdeclarative5-dev qtmultimedia5-dev \


petitlapin:
Thanks for testing and sorry for delay in answering,

libssl should not be required at compile time (it is an optional dependency at runtime if you need to download the assets) so cmake should not complain about it. Do you have the error?

AlexTheAussieNerd:
I've just re-run and can confirm the error was reported after running "cmake .." (on a raspberry pi 400, latest 64bit os)

This isnt exactly the error message as I'm typing from another screen, but the numbers will be identical.

The source was downloaded from a git clone https:// official mirror repository


The following REQUIRED packages have been found:
 * Qt <lots of packages>
 
- The following OPTIONAL packages have not been found 
- ECM (required version > 1.4.0)
-
- The following REQUIRED packaged have not been found:
 Open SSL

Cmake Error at /usr/share/cmake-3.25/Modules/FeatureSummary.cmake:464 (message: 
  feature_summary() Error: REQUIRED packages(s) are missiong, aborting CMake run.
Call Stack (most recent call first):
 CMakeLists.txt:143 (feature_summary)

- Configuring incomplete, errors occurred!
- See also "..GCompris-qt/build/CMakeFiles/CMakeOutput.log"  
- (CMakeOutput.log didnt show and error about it thought).

petitlapin:
You're right (https://invent.kde.org/education/gcompris/-/blob/master/CMakeLists.txt#L132-135), I'll add the dependency in the README when I have 5 minutes. Thank you!

HACKING
```diff
@@ -20,7 +20,7 @@ Checkout the source code by cloning it from GitHub
 If you want to have a personal fork, in order to commit occasional patches,
 fork it on GitHub then clone your fork
 
-    git clone git@github.com:<username>/GCompris-qt.git
```
petitlapin, line 23: this works if you have a ssh key (and this allows you to push without inputting your account/password everytime)

petitlapin:
I've just pushed it in https://invent.kde.org/education/gcompris/-/commit/2b7989ab54c461efb49a55c4fc9de7286c99a07b.

Thank you!

PR #294, opened 2024-01-24
Fix cmake qt5 plugins directory search for Archlinux

DameNoSupaplex:


petitlapin:
Thank you!
I'm also on Archlinux but never got the issue. But I see I have a bunch of symbolic links, maybe it is because of it...
The main repository is in https://invent.kde.org/education/gcompris/ so I've reported your commit in it directly: https://invent.kde.org/education/gcompris/-/commit/30d644c58e40a51f83d0ec297a985d0d04c0854c.

I'm closing the PR here. Thanks again!

PR #300, opened 2025-06-16
si

wildlux:
Hello this is mine first "pull request i want to translate Gcompris in Sicilian Language i hope i do well.
Thanks for any support to this amazing software!

Help is accept :-D


petitlapin:
Hi,
is there more we can do to help integrating this file?

